PR #4, opened 2023-10-31
Create the front-end navigation

RyotaMitaraiWeb:
This PR introduces some of the major implementations for the navigation menu of the application, in particular:
- a reusable logo component that can be used in any size
- a non-functional search field
- various buttons, such as a shopping cart button
- a mobile version of the navigation menu, which can be accessed starting from < 1024px viewport.

Some features that cannot be implemented currently, but should be when both projects progress enough to allow this:
- make the search field functional
- make the buttons functional
- replace all placeholder links with a client-side router's link components
- potentially improve the responsiveness of the navigation menu; the current version is usable down to roughly 560px, after which some components (particularly the logo and search field) may start to overlap.

Some of the components are also documented. The ones that are not are either non-functional and may be documented later on or, in the case of the mobile navigation menus, merely exist to prevent the main component(s) from becoming too huge on their own.

Here are some screenshots of the results:
nav_1
nav_2
nav_3
nav_4
nav_5
nav_6


julkascript:
Good job, here are a couple of things to improve:

- [ ] Remove the documentation READMEs.
- [ ] Name all directories in camelCase convention (for consistency sake).

I know you probably dedicated a lot of time to write the READMEs and I really appreciate that. However, I think that such documentation will bring extra complication, maintenance and most importantly - will reduce our flexibility. That's why we use typescript - there we have described all props and types, which helps with "documenting". I am okay with occasionally throwing in comments to explain stuff, but I believe the best way to go is to:
- use nice and descriptive names of the variables and functions in our code
- correct and fluent types
- write simple and readable code

frontend/src/components/Navigation/Desktop/Buttons/NewActionButton/NewActionButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+interface IMenuItem {
+  label: string;
+  href: string;
+}
+
+/**
+ * A simple button that opens a menu for a new action.
+ * @returns
+ */
+function NewActionButton(): JSX.Element {
```
julkascript, line 15: Move all buttons' files up in the `buttons` directory.
What we currently have is:
- `/Buttons/NewActionButton/NewActionButton.tsx`
- `/Buttons/SearchButton/SearchButton.tsx`
- ...

Move to:
- `/buttons/NewActionButton.tsx`
- `/buttons/SearchButton.tsx`
- ...

Creating subdirectories for each file is unnecessary and will likely bring complications.

julkascript:
Note: I appreciate your extra effort to make the navigation responsive, we will keep that, of course. Maybe I did not make it clear in the beginning but for the scope of our project (for 2-3 months of development), mobile responsiveness is not in scope, we are not aiming for that, we are only aiming for responsiveness for desktop. 

However, if you decide to also make your work responsive for mobile, I would appreciate this, and we will always include it.

RyotaMitaraiWeb:
- Removed READMEs. I have left the JSDocs comments in, but can remove them if necessary (I think they are convenient for the fact that they are displayed on hover in VSCode, but I can remove them if desired)
- changed all interfaces to types
- renamed folders to be camelCased (let me know if I missed something)
- in light of the button folder move, I also reorganized a bit the mobile navigation menu directory structure, since it suffered from a similar problem

Regarding the responsiveness: I will obviously always prioritize desktop, especially when we are running low on time, but in the cases where I have more than enough time to make it responsive (and I generally avoid going for super complex differences for mobile, meaning that responsive design shouldn't really take that much time), I think it makes sense to make stuff responsive, as this improves the project significantly, in my opinion.

julkascript:
Nice, though it appears after the latest commits that all directories are renamed to begin with a capital letter, though your imports look for the lowercase variants.

`import Navigation from './components/navigation/Navigation';` will not be valid, because your directories are still uppercased.

- [ ] Make all directories follow the `camelCase` convention (lowercase first letter and all letters after it, then capitalize only when separating for a word - example: myLongName).

RyotaMitaraiWeb:
after something like an hour or two of my files and folders getting locked left and right, it seems like my renames actually went through (apparently renaming them via VSCode's file explorer wasn't enough the first time... you really do learn something new every day)

julkascript (approved):

PR #8, opened 2023-11-09
Integrate React Router DOM into the project

RyotaMitaraiWeb:
This PR introduces React Router DOM to the project, which will enable client-side navigation. In addition to installing the library, this PR includes the following changes:

- a ``Router`` component which encapsulates the RouterProvider and its routes object. The reason for this is mostly so that the route object can be in a ``.tsx`` file (which improves the IntelliSense for the ``component`` property). In the event the routes object becomes too complex, it can possibly be separated in its own file. The current routes will have to be adjusted when more pages are developed and are there only for testing purposes.

- ``App`` is now rendered as a layout via the router and the ``main`` file renders the router itself.

- a configuration file that overrides various MUI components' linking mechanism. The ``linkBehaviorConfiguration`` file forces them to use the ``Link`` component from React Router DOM instead of the standard ``a`` tag when they are used as hyperlinks. The file was placed in the root of ``src`` since I wasn't sure whether it was a good idea to create a separate folder for such things.

- most of the existing codebase has been updated to use React Router DOM's features and received some preliminary endpoints (which obviously can change later on when needed)

- a simple 404 page that will be displayed every time an unhandled exception is thrown. The ``NotFound`` renders some of the ``App`` features (like the navigation and the MUI themes) for its own purposes since those do not show up in the component (as a matter of fact, the theming could be separated in its own object that can be imported anywhere, but I feel that this is outside of this PR's scope, so haven't touched that). An error boundary to catch the errors and do something (e.g. display a notification) could probably be implemented as to not display a 404 page for every unhandled exception, but this is likely best done in a different PR

in addition, I created a placeholder file for a folder called ``guards`` within the ``router`` folder, this is mostly to register the folder with Git. Not sure if there will be fetchers (or whatever the equivalent of Angular resolvers would be called here) or other similar features, so haven't created folders for those yet.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,7 +13,8 @@
         "@mui/icons-material": "^5.14.15",
         "@mui/material": "^5.14.15",
         "react": "^18.2.0",
-        "react-dom": "^18.2.0"
+        "react-dom": "^18.2.0",
+        "react-router-dom": "^6.18.0"
       },
       "devDependencies": {
         "@testing-library/react": "^14.0.0",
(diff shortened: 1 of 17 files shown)


julkascript:
Nicely done. I don't think we come to need a `guards` directory, but we will create one if it becomes necessary. Last thing I probably forgot to mention in the Jira task:

- [ ] Create a pages directory, where all pages' main files will be located.

Note: All main pages go to the `pages` directory, and all components to the `components` dir.

RyotaMitaraiWeb:
removed guards folder

regarding ``pages``, it seems like I accidentally put it in the ``components`` folder instead of the root of ``src``, this should hopefully be fixed now

julkascript (approved):

PR #9, opened 2023-11-12
Completing Dockerizing the backend

ZhivkoTringov:


backend/cardflow/settings.py
```diff
@@ -73,7 +73,7 @@
         "NAME": os.getenv("POSTGRES_DB"),
         "USER": os.getenv("POSTGRES_USER"),
         "PASSWORD": os.getenv("POSTGRES_PASSWORD"),
-        "HOST": os.getenv("POSTGRES_HOST"),
+        "HOST": "postgres",
```
julkascript, line 76: Use environment variable

backend/entrypoint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+echo "Waiting for postgres..."
+
+while ! nc -z 'postgres' 5432; do
```
julkascript, line 5: Use environment variables for the host and port

backend/entrypoint.sh
```diff
@@ -0,0 +1,23 @@
...
+  sleep 0.5
+done
+
+echo "PostgreSQL started"
+
+python manage.py flush --no-input
+python manage.py migrate
+
+
+if [ -z "$(python manage.py shell -c 'from django.contrib.auth.models import User; print(User.objects.filter(username="your_superuser_username").exists())')" ]; then
```
julkascript, line 16: `your_superuser_username` is a literal string. Make use of a real variable or an environment variable

PR #12, opened 2023-11-13
Backend/CF18 Authorization/Authentication

milensski:
#### Implementing authentication and authorization with Simple-JWT
 - Configured Simple JWT
 - added endpoints: `/accounts/api/login/` , `/accounts/api/register/` , `/accounts/api/refresh/`
 - added middleware for authorization of JWT tokens and handling the coresponding response - 401 if not OK

**Additional Info**

Backend server needs Encryption keys for the token encryption RS256
to configure this please generete the private and public keys and add them to `/keys` folder 
```
ssh-keygen -t rsa -b 4096 -m PEM -f jwtRS256.key
```

I will be happy to discuss it.


backend/cardflow/urls.py
```diff
@@ -3,9 +3,11 @@
 from drf_spectacular.views import SpectacularAPIView, SpectacularSwaggerView, SpectacularRedocView
 from rest_framework import permissions
 
+
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('api/schema/', SpectacularAPIView.as_view(), name='api-schema'),
     path('api/docs/', SpectacularSwaggerView.as_view(url_name='api-schema'), name='api-docs'),
     path('api/redoc/', SpectacularRedocView.as_view(url_name='api-schema'), name='api-redoc'),
+    path('accounts/', include('accounts.urls'),)
```
julkascript, line 12: Ensure that the entrypoint is `api`, and then your path. `/api/my/new/path`

julkascript:
Exclude the admin as an authorized endpoint

julkascript:
I cannot hit the register endpoint. I am trying:
POST `/accounts/api/register/` with JSON data containing the username and password but I get 401.

I believe you have to also allow the register endpoint to end in a slash '/' and see how the logic goes. The `register` endpoint should be hit without tokens and should not be an authenticated endpoint.

julkascript:
I also get 500 error when trying to hit POST `/accounts/api/login/` with JSON data containing the username and password.

Error: 
```
django.http.request.RawPostDataException: You cannot access body after reading from request's data stream
```

julkascript:
> The issues should have now be fixed. I have tested with the provided parameters and it is working. All endpoints are reachable in the excluded list.
> 
> Sorry for the inconvinience with the endpoints.

Good job, it looks very good. One thing is still missing - I cannot access the admin page. I try `http://127.0.0.1:8000/admin/`. Ensure that this is opened to all users. Also, ensure all actions in the admin are also working (e.g. logout).

milensski:
> > 
> 
> Good job, it looks very good. One thing is still missing - I cannot access the admin page. I try `http://127.0.0.1:8000/admin/`. Ensure that this is opened to all users. Also, ensure all actions in the admin are also working (e.g. logout).

That's very strange, I had tested the admin as well, but will look into it and verify everything again. Maybe I have missed something.

milensski:
> 
> Good job, it looks very good. One thing is still missing - I cannot access the admin page. I try `http://127.0.0.1:8000/admin/`. Ensure that this is opened to all users. Also, ensure all actions in the admin are also working (e.g. logout).

The admin is now is working properly all endpoints are reachable and I have added the swagger endpoints as well if we will use it.

julkascript (approved):

PR #13, opened 2023-11-15
Create static profile pages

RyotaMitaraiWeb:
This PR introduces two profile pages: one for the settings of the user and one that displays public data about them.

The pages are currently mostly static; they do not retrieve data about the user in any way and the data there is hardcoded. The only functionality present is that the settings page applies some minor validations to the inputs before enabling/disabling the corresponding buttons, but those can change as needed

Some keypoints:
- In the settings page, I have made it so that the upload happens with a click on the avatar circle. I also added some clarifying text due to this. The file size limit (currently 2MB) and the supported image files (currently whatever the browser's file explorer shows for this category) can be changed if needed, those have been left as they are for demonstration purposes.

- there is a placeholder page for the blog endpoint. This should ideally be removed in the future when that page is ready or scrapped.

- I created some general-purpose components (e.g. a "Page section" component) to make my work easier and to avoid having to refactor it later on; merging this PR will also make them usable for future pages.

I tried to make the directory structure organized, but I expect there to be room for improvements and am willing to restructure it if requested. The ``profile`` directory in ``components`` holds all components that are later used in the ``pages`` components to keep the ``pages`` folder as tidy as possible.

edit: I had to change the theming for the warning color, this shouldn't really affect anything existing as far as I am aware, but just to note it


julkascript:
I advise you to simplify the structure. The directory nesting in some places is okay - when we have a group of files related to one and the same thing. However, I think we can simplify like:
 ```
PageSection.tsx <-- out of its directory
PageHeader.tsx <-- out of its directory
 profile/
	ProfileNavigation.tsx <-- out of its directory
	ProfilePage.tsx <-- out
	ProfileSectionFooter <-- out
	
	profileSettings/
			DeleteAccount.tsx <-- out
			EmailSettings.tsx <-- out
			ShipmentAddressSettings.tsx <-- out
 ```

julkascript:
I created a color palette in the Figma file. You can have a look and try to use as much of the styles as possible.

frontend/src/components/profile/profileSettings/deleteAccount/DeleteAccount.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <div className="pt-4 pb-4 lg:pl-12 lg:pr-12">
+        <h2 className="font-bold mb-4 text-lg">Delete your account</h2>
+        <p>
+          Permanently remove your Cardflow account. Please proceed with caution, this action is
+          irreversible.
+        </p>
+      </div>
+      <ProfileSectionFooter backgroundColor="red">
+        <p>We will definitely miss you...</p>
+        <Button color="warning" variant="contained" className="inline-block">
```
julkascript, line 17: Use error color from the color palette instead

frontend/src/components/profile/profileSettings/avatarSettings/handleAvatarUpload.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Converts the passed file to a Base64 string while applying some
+ * validations to it.
+ * @param file
+ * @returns a Promise that resolves to a Base64 string. If the uploaded
+ * file is not an image, is over 2MB, or the upload fails for any reason,
+ * this function will reject with a corresponding error message (a string).
+ */
+export const handleAvatarUpload = async (file: File) =>
```
julkascript, line 9: The upload works very well, good job! However, I noticed one issue. When you have already uploaded a file, and you go to browse, if you just select a file but close the modal, instead of uploading - it crashes with errors in the console:
image

However, you do not have to fix this issue now. I will add it as an `issue` to the issues list in the repo.

frontend/src/components/profile/profileSectionFooter/ProfileSectionFooter.tsx
```diff
@@ -0,0 +1,34 @@
+import { Typography } from '@mui/material';
+
+type color = 'gray' | 'red';
+
+type ProfileSectionFooterProps = {
```
julkascript, line 5: Since the `Delete account` section requires a red background color and a red border, I would either advise you to add a border prop.

julkascript:
When visiting `/user/<username>`, the Public Info tab in the sidebar should be the active item selected by default.

RyotaMitaraiWeb:
All of these should be addressed with the following notes:

- the upload issue seems to be a Chromium thing only (no idea about Safari since I don't use Apple stuff), since I could not reproduce it on Firefox (which is the browser I use mainly and also probably why this bug went through unnoticed... I should probably test a bit more thoroughly in other browsers from now on). Fortunately, I quickly identified what was causing it, so it should hopefully be addressed now

- Based on what I saw in the Figma wireframes, the border colors seem to apply for the entire section and not just the footer, so I played around and made it possible to specify the colors for both (though I have to note that the red border is barely visible, which is probably why I missed it the first time), hopefully that's what was being requested.

- the navigation issue was caused by the hardcoded links. Those now are generated by the ``username`` path in the URL. The data in the profile links is still hardcoded, though.

if I missed anything, please let me know. Also, it is possible for some parts of the app to have been missed during the theme updating, so let me know if you catch anything I missed in that regard.

julkascript:
- [ ] Resolve conflicts

julkascript (approved):

Issue #14, opened 2023-11-17
Superuser not created on container run


julkascript:
The command for creating a superuser in the `entrypoint.sh` file is not working, it needs to be fixed. Fix it and create a PR.

PR #15, opened 2023-11-18
Fix issue #14: Superuser not created on container run

milensski:
## Changes made:

1. Changed the outer double quotes to single quotes in the `if condition` to avoid issues with nested double quotes.
2. Changed the import of the `User` model - since the default User model has been swapped with a custom User model in the `accounts` app which was leading to error ->
``` 
AttributeError: Manager isn't available; 'auth.User' has been swapped for 'accounts.User' 
```
3. Added the email field when creating the super user 
4. Added *email* and *password* fields to be used from the environment variables - I think it is more convinient to use it that way.

### DIFF
--- a/backend/entrypoint.sh
+++ b/backend/entrypoint.sh
@@ -18,9 +18,9 @@ echo "PostgreSQL started"
 python manage.py flush --no-input
 python manage.py migrate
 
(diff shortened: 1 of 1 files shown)


julkascript:
Closes #14

julkascript (approved):

PR #16, opened 2023-11-21
Implement authentication and authorization on the front end

RyotaMitaraiWeb:
(disclaimer: slightly long PR)

this PR introduces various tools to assist with authentication and authorization later on. Before I begin, I will note that there are a few instances where data (more specifically, the user_id) is extracted from the access token provided by the server, since there doesn't seem any other way to obtain data about the currently logged in user. If this is an issue and there are plans to support an alternative way to obtain such data, I will refactor it.

the following features have been implemented here:
- an HTTP service that makes requests with some preconfigured options. The purpose of this service is to be used in the implementation of other services. The service itself is actually one single function and all its methods are just calling the function for the different HTTP methods. The reason I have opted for this is mostly to improve readability, though I personally don't have an issue refactoring it to be just that one function. The service also implements the refresh token requirement; when the server responds with a 401 to the initial request, the service attempts to retrieve a new access token and then redo the request. I am using the status code here because the only indicator of an expired token per the server is the error message, which I think is a very shaky way of determining whether the token has expired on the client.
- - **Note:** the service comes with a few unit tests that mostly test implementation. The refresh token requirements are not tested because trying to mock consecutive calls in Vitest is pretty cumbersome and I didn't feel it was worth it. From local testings, it seems to work, though more eyes on that would certainly be helpful
- a user service that handles the user's data. I have implemented simple register and login methods as well which can be used in CF-17.
- a hook called ``useCurrentUser`` which is a context hook that provides data about the currently logged in user and two methods to manipulate it. There is also ``useAuthenticationStatus`` which internally uses the aforementioned hook to provide data about the user's authentication status in a more readable manner.
- two components, ``Authorized`` and ``Unauthorized``, which render their children only when the user is logged in or not logged in, respectively
- an ``authorizeGuard``, which can be used as a ``loader`` to protect pages from unauthorized users
- I have also implemented session checking upon app load

I have also created myself a constants folder where I can put things like API endpoints cause I find it more comfortable to have one place for all of them rather than have magic strings all over the project.

I am open to refactoring all of the above if necessary


RyotaMitaraiWeb:
PR has been updated with the new claims in the JWTs. In addition, the profile page settings are now prefilled with the email and username (I have also made the save buttons disabled if the value is unchanged to prevent unnecessary requests to the server) and the settings link in the profile navigation menu is hidden from the other users.

julkascript:
Good job with the PR. I like the `httpService`. It's not a big improvement, but let's move the tokens to cookies, instead of `localStorage`. It's generally considered to be "safer" to store the tokens in cookies. Local storage makes the tokens vulnerable to XSS attacks.

frontend/src/App.tsx
```diff
@@ -44,6 +47,22 @@ function App() {
     [],
   );
 
+  const { setUser, restartUser } = useCurrentUser();
+
+  useEffect(() => {
+    userService
+      .verifySession(localStorage.getItem('refreshToken'))
```
julkascript, line 54: Don't send requests if there is no `refreshToken`.

RyotaMitaraiWeb:
``App.tsx`` (and ``authorizeGuard`` since the logic is arguably applicable there too) now considers the user unauthenticated if either token is missing

regarding localStorage vs cookies:

I am not a cybersecurity expert by any means, but as far as I am aware, cookies help against XSS attacks only if they are HTTP-only. However, since the API seems to require Authorization headers for the access token and a request body for the refresh token, this means I'd have to store the tokens in non-HTTP only cookies in order to attach them properly (so theoritically, if you suffer an XSS attack, you could read those cookies just as easily as with localStorage items). There is no improvement as far as I am concerned and is arguably more of an inconvenience, since localStorage at least has an API to make it more comfortable to work with.

I *could* implement this, but I don't think security will be a benefit in our case, unless I am missing something (in which case I am happy to be corrected)

julkascript:
> I am not a cybersecurity expert by any means, but as far as I am aware, cookies help against XSS attacks only if they are HTTP-only. However, since the API seems to require Authorization headers for the access token and a request body for the refresh token, this means I'd have to store the tokens in non-HTTP only cookies in order to attach them properly (so theoritically, if you suffer an XSS attack, you could read those cookies just as easily as localStorage items). There is no improvement as far as I am concerned and is arguably more of an inconvenience, since localStorage at least has an API to make it more comfortable to work with.
> 
> I _could_ implement this, but I don't think security will be a benefit in our case, unless I am missing something (in which case I am happy to be corrected)

Agreed, we will keep the `localStorage`.

frontend/src/services/http/http.ts
```diff
@@ -0,0 +1,103 @@
...
+export const httpService = {
+  get<TResponseBody>(url: string) {
+    return request<TResponseBody>('GET', url);
+  },
+  post<TResponseBody>(url: string, body?: any) {
+    return request<TResponseBody>('POST', url, body);
+  },
+  put<TResponseBody>(url: string, body?: any) {
+    return request<TResponseBody>('PUT', url, body);
+  },
```
julkascript, line 28: We also need a `patch` method.

RyotaMitaraiWeb: PATCH has been added to the HTTP service

julkascript:
LGTM

julkascript (approved):

Issue #21, opened 2023-11-27
Docker not reacting to changes in the code


julkascript:
Have a code change and observe how the Docker server is not picking up the change.

dinocom33:
To make Docker responsive to code changes, we added the following line to the docker-compose.yml file:
```
    volumes:
      - .:/usr/src/app
```
We also found that when the container is restarted, the base is completely cleared by the following command in entrypoint.sh file:
```
python manage.py flush --no-input
```
As we are not sure if this is a bug or by design, this command has only been commented out and not removed.

PR #22, opened 2023-11-28
Fixes issue #21

dinocom33:
Fixed issue#21

### DIFF
--- a/backend/accounts/middleware.py
+++ b/backend/accounts/middleware.py
@@ -25,7 +25,7 @@ class JWTAuthorizationMiddleware:
 
     def __call__(self, request):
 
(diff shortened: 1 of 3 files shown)


backend/accounts/middleware.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, get_response):
 
     def __call__(self, request):
 
-        if request.path in self.EXCLUDED_PATHS or request.path.startswith('/admin'):
+        if request.path in self.EXCLUDED_PATHS or request.path.startswith('/admin') or request.path.startswith('/api/yugioh'):
```
julkascript, line 28: It''s not a problem, but try not to have changes that are not related to this PR, especially at work.

julkascript (approved):

Issue #18, opened 2023-11-25
`/api/docs` not accessible without a token


julkascript:
`/api/docs` not accessible without a token
`/api/docs/` is accessible without a token

Enable it as a path.

PR #23, opened 2023-11-29
Disabling custom middleware related to fix issue #18

milensski:
I have disabled the custom middleware since it is a bit overkill at this moment, while the standard JWT middleware is doing the same work and we can take advantage from the permission classes for controlling the access to the endpoints without excluding them in a list one by one.

It is just commented it in `settings.py` so we can have it for later on, if needed.

Please let me know if this fixes the issue for now, or should I re-thing the logic of the custom middleware and fix it there.


julkascript:
Viewsets will use `rest_framework.permissions.IsAuthenticated` to authorize users. 
@milensski @dinocom33

PR #24, opened 2023-12-02
Backend/cf19

dinocom33:
This PR represents CRUD operations for listing(card for sale). Users can add/edit/delete their own listings only.
A new endpoint has also been created through which the frontend can search for listings based on listed or unlisted.
The response looks like the following:
```
[
  {
    "id": 14,
    "card": 171,
    "card_name": "Fantastical Dragon Phantazmay",
    "card_set_id": 106,
    "user": 2,
    "user_name": "simo",
    "price": 502,
    "condition": "mint",
    "quantity": 1,
    "is_listed": false,
    "is_sold": false
  },
  {
    "id": 17,
    "card": 15,
    "card_name": "Anti-Spell Fragrance",
    "card_set_id": 13,
    "user": 3,
    "user_name": "simo1",
    "price": 150,
    "condition": "near_mint",
    "quantity": 1,
    "is_listed": false,
    "is_sold": true
  },
  {
    "id": 18,
    "card": 2,
    "card_name": "Accesscode Talker",
    "card_set_id": 1,
    "user": 3,
    "user_name": "simo1",
    "price": 200,
    "condition": "mint",
    "quantity": 1,
    "is_listed": false,
    "is_sold": false
  }
]
```


backend/listing/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path, include
+from rest_framework.routers import DefaultRouter
+from . import views
+
+router = DefaultRouter()
```
julkascript, line 5: No need to have two Viewsets that get the same job done, use just one. Also, no need to add to the `listings` name. Having a simpler paths in the API makes it easier to use and understand.

Desired API paths:
`/api/listing`, `/api/listing/1/`

julkascript: @milensski you might have to change your implementation afterwards.

dinocom33: - To perform listing search for all users use base endpoint.
- To perform listing search for specific user use endpoint with ?user_id=<user_id> parameter.
- To perform PUT or PATCH use endpoint with /<listing_id> parameter.

julkascript (approved):

PR #25, opened 2023-12-04
Backend/cf-24/Buy functionallity

milensski:
In this PR I have implemented Buy functionallity to the ListingView endpoint `http://localhost:8000/api/listing/yugioh_listing/{id}/` with both PUT and PATCH methods. 
- When the field **is_sold** is changed to *True* - the listing becomes "unlisted" and cannot be sold more than once or changed.
- If the listing is unlisted it cannot be sold untill it is made listed.
- I have added some extra checking for the **is_listed** value to avoid the scenario where the listing is sold but not unlisted

>[!IMPORTANT]
>This implementation is based on the branch with PR #24


backend/listing/views.py
```diff
@@ -0,0 +1,101 @@
...
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = ListingFilter
+
+    def get_queryset(self):
+        return self.queryset.filter(user=self.request.user)
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
+
+    def update(self, request, *args, **kwargs):
```
julkascript, line 32: The `update` endpoint should be dedicated to just generic fields updating.

The `buy` functionality deserves its own endpoint because it is a different type of logic. In the future, the buy functionality might trigger other processes, like creating an order, sending emails, etc. If this is in the `update` endpoint, you can imagine how bad it will get eventually, so that's why it would be ideal if we can put it in a separate endpoint.

Do:
`/api/listing/<id>/buy`

Don't:
`/api/listing/buy?id=2`

milensski: Notes taken,
I have re-implemented the logic following the suggestions

I have seperated the functionallity to 3 endpoints:
**PUT** request to - `/api/listing/<id>/buy/` - will update the listing as sold and make it unlisted
**GET** request to - `/api/listing/sold/` - will return all *sold* listings
**GET** request to - `/api/listing/unsold/` - will return all *unsold* listings

>[!Important]
>I have made а custom permission class to check if the user is the same as the owner which is trying to mark the listing as sold. This can be used for any of the other ViewSets as well.

Please let me know if I need to change the quering for sold and unsold listings as a query parameter to the `/api/listing/?is_sold=true` endpoint. I was not sure which way to choose. 

Also I used the PUT method for this endpoint `/api/listing/<id>/buy/` because I think it is more intuitive since it updates an existing object rather than creating new object which in my opinion is a job for the POST method 🙂

julkascript: I would generally go for a query parameter, since you only filter the data, and you don't do any more actions. However, it's okay for now, no need to fix it at this point.

milensski: Okay, thank you for the feedback, I will wait for the listing task to be ready and I will correct this accordingly.

backend/listing/views.py
```diff
@@ -0,0 +1,105 @@
...
+        - API endpoint that query for unsold listings \n
+    """
+
+    queryset = Listing.objects.filter()
+    serializer_class = ListingSerializer
+    permission_classes = [permissions.IsAuthenticated]
+
+    def get_permissions(self):
+        if self.action == 'mark_as_sold':
+            return [IsOwner()]
```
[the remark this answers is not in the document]
julkascript: You are right, but with this implementation it turns out a user can only buy his listing. Other users cannot buy his stuff.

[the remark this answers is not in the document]
milensski: I have now removed this method and the viewset now is for authenticated users. It appears my logic was wrong.
Also I added *if the user is the owner* **not** to be able to buy his own listing.
Please let me know if now is okay.

julkascript (approved):

PR #26, opened 2023-12-14
Create details page

RyotaMitaraiWeb:
This PR introduces the card details page. The page fetches and passes data from the backend where possible. The places where data is hardcoded is as follows:
- the "Available items", "Number", "From", and "30-days average price" cells are hardcoded, as no backend endpoint provides them or I don't know how I would generate them
- the entire table at the bottom of the page is hardcoded at the moment for two reasons, the first one being that there doesn't seem to be an endpoint that allows me to get all listings by the ID of the page (the ``listing`` endpoint gives *all* the listings and I'd have to filter them on the client, which is not a good idea for a multitude of reasons) and the second one being that the listing GET endpoints are not available to unauthenticated users (CF-20 says that the listings in the page should be visible even to unauthenticated users, so I am not sure if those are appropriate for this scenario or this is the result of an error). Let me know if I got something wrong

In addition, I have also done the following:
- I updated the httpService to allow query string construction. In other words, you can now pass an object as the last argument of each method and this will append all key-value pairs as query strings appropriately. This was done in this PR since I was expecting to need something like this, but it doesn't seem like I got to use it here.
- I also added the star badge to the profile page for demonstration purposes. Small note: the public profile page treats the golden star as an achievement for >10K sales, but other wireframes treat it as an achievement for >5K sales, not sure which one is correct.

**Note:** in the hardcoded listings table, the UK flag image is fetched via hotlinking. This is only for demonstration purposes and will probably change to something more reliable when possible.

**Edit:** one of the types wasn't 100% correct (more specifically, the ``YugiohCard`` one), this was because I used Swagger's schemas when I was creating them, but it turns out that the returned data doesn't match the schema entirely. I have committed a fix for this.


julkascript:
Raised issue #29 to fix the GET listing for unauthenticated users.

frontend/src/components/yugioh/YugiohCardImage.tsx
```diff
@@ -0,0 +1,20 @@
...
+  src: string;
+};
+
+function YugiohCardImage(props: YugiohCardImageProps): JSX.Element {
+  const theme = useTheme();
+  const secondary = theme.palette.secondary.main;
+
+  return (
+    <div
+      className={`bg-white p-8 w-[314px] h-[422px] border-[${secondary}] border rounded flex justify-center items-center`}
```
julkascript, line 13: `rounded` -> `rounded-lg` to match the table

frontend/src/components/yugioh/table/market/AddToCardButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+        backgroundColor: '#15B58D',
+        borderColor: '#3A3A3A',
+        color: 'white',
+        ':hover': {
+          color: 'black',
+        },
+      }}
+      aria-label="Add to cart"
+      size="small"
+      className="border h-12 w-14"
```
julkascript, line 18: - [ ] `AddToCard` -> `AddToCart`
- [ ] Make the height of the button `h-10`

julkascript:
> Raised issue #29 to fix the GET listing for unauthenticated users.

We will provide pagination in a further story. It's okay for now.

julkascript:
Overall, great job with the PR, love it, super minor comments only.

frontend/src/components/yugioh/YugiohCardImage.tsx
RyotaMitaraiWeb: This is done

frontend/src/components/yugioh/table/market/AddToCardButton.tsx
RyotaMitaraiWeb: This is done

julkascript:
> * Raised issue [GET listing endpoints are not available to unauthenticated users #29](https://github.com/julkascript/cardflow/issues/29) to fix the GET listing for unauthenticated users.
> * Pagination requested via [GET listings endpoint to have a pagination #30](https://github.com/julkascript/cardflow/issues/30)

Issues delivered. See the pagination [docs](https://www.django-rest-framework.org/api-guide/pagination/).

RyotaMitaraiWeb:
In order for this to be "working", I will need the backend to provide me with an endpoint (be it an existing or a new one) that allows me to get all listings with a given ``card_set_id``, as I have no ways to even filter it on the client due to the pagination.

I didn't communicate this well earlier, for which I apologize. I have raised a request for this ( #32). On the bright side, I tried to abstract away most of the logic here and add/update some unit tests, so hopefully there should be minimal, if any, edits needed if my request is implemented

RyotaMitaraiWeb:
integrated the card_in_set_id query into this PR (considering that it seems to be working) and added basic pagination. The pagination in this case doesn't use the current URL (thus loading the page will always start at page 1), but if that's desired, let me know.

**Note:** the flag column is left for now, but seeing as it's currently hardcoded, I could probably remove it if desired.

julkascript (approved):

PR #28, opened 2023-12-17
Add semi-functional search

RyotaMitaraiWeb:
This PR makes the global search field functional and also introduces a semi-functional search page.

When the user types on the global search field, a debounced request is made and up to four results are displayed. When the user clicks on one of the results, they are redirected to its page. When the user clicks the search icon button or the "Show all" results, they will be redirected to the search page.

Several things need to be mentioned:
- the search page uses dynamic routes rather than query strings as React Router DOM's loaders cannot access query strings
- the search page does not have pagination and will display all results on one page, I can potentially implement this at a later stage (ideally, we'd have pagination for that purpose)
- the global search field merely redirects the user when they want all results. Because the search page itself sends a request when loaded, this means that the client effectively makes two requests. I can potentially implement some sort of caching to prevent this at a later point
- The filter has not been implemented, because most (if not none) of the filter categories aren't searchable by at this point
- The table does not have every single cell from the wireframes since those are not included with the cards endpoint. In addition, it is not stated how the colors for the rarity badge are determined (and since there seem to be over 20 different types of rarity, it's also impossible to derive this logically)

**Note:** this PR depends on features implemented in #26, I will update this whenever the need arises


julkascript:
- The filter will be out of scope for this task
- Rarity colors are out of scope, it's good enough the way it is implemented now

julkascript:
Looks good, works good, nice!

However, I found a small bug. When you switch between card details from the search, the listings are not changing to the new item's. We need to fix that.

How to reproduce:
1. Search for any card and open its details page.
2. Without leaving the page, search for another card and open its details page -----> observe how the new listings are not fetched.

julkascript:
> This should be fixed (there was a similar bug in regards to the search page, which I have also fixed)
> 
> For this to be complete, I will likely need the `/api/yugioh/cards_in_set/` endpoint to allow me to search cards by name (much like the standard cards endpoint), as proper pagination is not possible with the current setup and even the global search field underreports the total amount of results (as far as I have tinkered with Swagger). I will open a new issue after the holidays regarding this (as to not disturb the backend team)

You can open the issue and describe in detail what you need. The guys will pick it up whenever they can after the holidays.

RyotaMitaraiWeb:
refactored the PR with the new query (alongside a few minor design tweaks)

julkascript:
LGTM

julkascript (approved):

PR #39, opened 2023-12-29
Creating FrontEnd Workflow for running the tests

milensski:
I have created the GitHub Action workflow file for running the tests for the frontend

It is configured for all branches to run on pushes and pull requests, let me know with what settings to configure it if this is not suitable.


.github/workflows/action-test-frontend.yml
```diff
@@ -0,0 +1,42 @@
+name: Frontend Tests
+
+on:
```
julkascript, line 3: Sorry for not putting it as a requirement in the Jira, but we would ideally want the tests to be run only on PRs to develop, master.

frontend/package.json
```diff
@@ -23,6 +23,7 @@
     "react-router-dom": "^6.18.0"
   },
   "devDependencies": {
+    "@actions/core": "^1.10.1",
```
julkascript, line 26: I don't think this package is needed (unless I miss something)?

milensski: I have removed it now, 
it was for me to locally test the workflow but it is not needed as I see I can test directly on the branch in github

.github/workflows/action-test-frontend.yml
milensski: it is now changed to run only on PR on develop and master

julkascript (approved):

PR #40, opened 2024-01-01
Cf 26/be workflow

dinocom33:
This PR represents Github workflow for backend.

IMPORTANT NOTICE:

A change in settings file was to be made about database host, because Gitib Workflow is looking for a db connection to localchost address. With the change will first look for the host in the .env file and if does not find, by default will take "postgres":

from "HOST": "postgres"
то "HOST": os.getenv("POSTGRES_HOST", 'postgres')


julkascript:
This works, but why did you not opt to using docker directly, instead of manually recreating all the steps?

julkascript (approved):

PR #41, opened 2024-01-08
Create sell management page

RyotaMitaraiWeb:
This PR implements the sell management page, which is entirely functional. There are a few notes, though:
- the pagination differs from the wireframe for consistency with the other pages that use similar tables and to save time
- the "Delist all" and "Deselect all" buttons will perform the respective operations on all listings on the current page, regardless of whether they've been selected or not. The third button opens more options, which includes performing the actions on the selected items. Some of those options are also disabled when appropriate.
- since the branch for CF-20 is not published at the time this PR was opened, I resorted to creating the CRUD service methods I needed in this PR, this may result in a conflict or two (due to this, I've also opted not to include unit tests for the time being). Likewise, client URLs are not included / may not match with the ones used there and will require updating at some point

### DIFF
--- a/frontend/src/components/PageHeader.tsx
+++ b/frontend/src/components/PageHeader.tsx
@@ -25,10 +25,10 @@ function PageHeader(props: PageHeaderProps): JSX.Element {
 
   return (
     <header
-      className={`lg:pl-24 flex justify-between flex-row pb-8 pt-8 border-b-2 border-stone-300 text-center lg:text-left bg-white ${className}`}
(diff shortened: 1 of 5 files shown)


julkascript:
Good job. About conflicts with other PRs - the early bird gets the worm. :)

Regarding your PR, let's improve some UI stuff:
- [ ] Make the 3 dots of the helper menu thicker/bolder
- [ ] Make some space between the table and the top bar
- [ ] Add a "+" to the New Listing button
- [ ] Make all buttons and borders on this page `rounded-md`

RyotaMitaraiWeb:
did those + added white background to the menu (I didn't see it initally, whoops) + refactored a bit the page change when deleting items (so that it remains on the same page when possible)

julkascript (approved):

PR #42, opened 2024-01-11
Sell/newlisting

kthristow:


julkascript:
To do:
- [ ] Use Material UI icons, instead of the static icons - use icons as close as possible to the provided in Figma, doesn't have to be exact
- [ ] The `Cancel` button in the listing creation screen should return the user to the Sell page (`/sell/manage/`)
- [ ]  The `Post` button should only be clickable after the user has selected a card, quantity, price and condition (otherwise - disabled)
- [ ] After creating a listing successfully (Post button), the user should be returned to `/sell/manage/`. He should be able to see his new listing.
- [ ] Posting with the `Post another` option should post the listing and return the user to the cleared state, where he can select a new card and continue to post
- [ ] The user should be able to go to his listing in edit mode when clicking on his listing from the Sell page (`/sell/manage/`).
- [ ] There should be an edit mode when you open a listing that you own
- [ ] You should not be able to edit other users' listings (not even see the screen - edit mode)
- [ ] Move all sell logic under the `/sell` path (`/listing/newlisting/` -> `/sell/new/` (which already exists btw), 
- [ ] Since you created the Buy-Sell-Trade navigation - make it work for Buy and Sell (Buy leads you to `/search/`, Sell leads you to `/sell/manage/`)
- [ ] Make the Post button a main action button (not outlined but filled)
- [ ] Fix the paddings across the page and bolden the title (see pictures below)

image
image

frontend/src/components/sellListing/NewListingTopBar.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState } from 'react';
```
julkascript, line 1: There is already a component for this. Look for the <PageHeader> and make use of it, instead of re-implementing it.

julkascript:
The changes look good. There's more work to do:
- [ ] New requirement: "Clear" button in listing edit mode should not be visible 
- [ ] Clear button in listing create mode should clear all the inputs and return the user to the state where they started from (where they select a card)
- [ ] The URL of the edit mode should either be `/sell/listing/${id}/edit` or `/sell/listing/${id}`

frontend/src/pages/listing/SellListing.tsx
```diff
@@ -0,0 +1,271 @@
...
+    user: 0,
+    card_name: '',
+    card_set_id: 0,
+  });
+
+  useEffect(() => {
+    async function loadCardListing() {
+      const currentCard: YugiohCardListing = await yugiohService.getListingById(Number(params.id));
+      if (currentCard && currentCard !== null) {
+        setFormData(() => ({
```
julkascript, line 48: Use destructuring, instead of manual prop assignment. If there were 120 props, would you go for manually assigning them all?

frontend/src/router/Router.tsx
```diff
@@ -90,6 +93,24 @@ const routes = createBrowserRouter([
             element: <SellManagement />,
             loader: authorizedGuard,
           },
+          {
+            path: 'listing',
```
julkascript, line 97: - [ ] Remove outdated route

frontend/src/components/sellListing/NewListingTopBar.tsx
julkascript: @kthristow  The component is called <PageHeader />. Please re-use it. It has exposed header and children props.

frontend/src/components/sellListing/NewListingTopBar.tsx
```diff
@@ -0,0 +1,79 @@
...
+    condition === undefined;
+  card === null || card === undefined;
+
+  const handleCheckboxChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setPostAnother(e.target.checked);
+  };
+
+  const onSubmit = (e: React.FormEvent) => {
+    handleSubmit(e, postAnother);
+    if (!postAnother || postAnother === null) {
```
julkascript, line 38: `null` evaluates to false, so `!postAnother` is enough for the check

frontend/src/pages/listing/SellListing.tsx
```diff
@@ -0,0 +1,271 @@
...
+                  <select
+                    id="condition"
+                    name="condition"
+                    value={formData.condition}
+                    onChange={handleChange}
+                    className="w-28 px-2 py-0.5 text-sm text-left text-gray-700 border-none focus:ring-0 focus:outline-none"
+                  >
+                    <option value="poor">Poor</option>
+                    <option value="good">Good</option>
+                    <option value="played">Played</option>
```
julkascript, line 231: The `Good` quality is better than `Played`, so therefore, it should be after it.
`Poor -> Played -> Good -> Excellent`

julkascript (approved):

PR #43, opened 2024-01-15
Backend/cf 43

dinocom33:
This PR represents CRUD operations for user profile and the corresponding tests.

Two new libraries were installed and added to requirements.txt:
  - Pillow - adds image processing capabilities
  - django-cleanup - automatically delete old images when user changes his avatar


backend/accounts/urls.py
```diff
@@ -2,16 +2,18 @@
...
 app_name = 'accounts'
 
 router.register('register', RegistrationView, basename='register')
+# router.register('update', UserUpdateView, basename='account_update')
 
 urlpatterns = [
     path('', include(router.urls)),
     path('login/', MyTokenObtainPairView.as_view(), name='login'),
     path('refresh/', TokenRefreshView.as_view(), name='refresh'),
+    path('update/<int:pk>/', UserUpdateView.as_view(), name='account_update'),
```
julkascript, line 18: To me the name `update` does not make sense, because I can do more than update. I propose you to change it to just `user` because you do full CRUD. 

Everything else is very well done!

julkascript (approved):

PR #46, opened 2024-01-19
Allow the user to make the changes to their profile on the front end

RyotaMitaraiWeb:
This PR makes the profile settings page functional; all fields perform the necessary edits/deletions. A bunch of notes:
- when the user changes their username or email, they will be logged out. The reason for this is that their current tokens reflect their old username / email and I figured it'd be more appropriate to have them reauthenticate to receive tokens with up-to-date data. Corresponding texts in the fields have also been changed to reflect that. Let me know if this is not a desired behavior
- the httpService has been updated to handle multipart form data requests.
- the user context and session verification have been updated in this PR; when the user loads the app and has a valid session, the front end will make a request to retrieve the user data as well. This is because some of the newly implemented user data is needed in other parts of the app (for example, the avatar is needed in the navigation menu, the shipment address is needed in the upcoming shopping cart, etc.) and this prevents the need for the user data to be requested constantly.

EDIT:
- I updated the gitignore to include the directory with the avatars, let me know if this is not desired
- one of the type refactorings I did caused a build error in the Home page. This build error involved authentication logic that is already handled by ``App.tsx`` so I simply removed that part of the code. I mention this just so it's clear why I deleted it in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ dist-ssr
 *.njsproj
 *.sln
 *.sw?
+
+# Avatars
+backend/media
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -59,7 +59,7 @@ function App() {
         .then(async (jwt) => {
           localStorage.setItem('accessToken', jwt);
           const { user_id } = userService.extractUserFromToken(jwt);
-          const user = await userService.getUser(user_id);
+          const user = await userService.getUserById(user_id);
 
           setUser({ user_id, ...user });
         })
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -56,10 +56,12 @@ function App() {
     if (refreshToken && accessToken) {
       userService
         .verifySession(refreshToken)
-        .then((jwt) => {
-          const user = userService.extractUserFromToken(jwt);
-          setUser(user);
+        .then(async (jwt) => {
           localStorage.setItem('accessToken', jwt);
+          const { user_id } = userService.extractUserFromToken(jwt);
+          const user = await userService.getUser(user_id);
+
+          setUser({ user_id, ...user });
         })
         .catch((res) => {
           if (res instanceof HttpError && res.err.status < 500) {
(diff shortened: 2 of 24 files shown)


julkascript:
Your additions and personal touches are all good, I approve them. However, there is one small bug I am experiencing at my side:
- when I log in or sign up, I am not logged in right away, but I need to refresh (tokens are in the storage, though)

Also, sorry if it was missed in the Jira task, but could you also do the `Public info` part of the profile? - it would be enough to do the first section's username (e.g. @ivan.dimitrovv) and public image.

RyotaMitaraiWeb:
It looks like the code that I deleted was actually handling the authentication after the user logs in or registers successfully, which is what brought the regression (I probably should have checked more carefully, since I was under the impression that the authentication happened within those components and not in the home page). I changed the auth forms components a bit so that they are the ones to authenticate the user

RyotaMitaraiWeb:
> Also, sorry if it was missed in the Jira task, but could you also do the `Public info` part of the profile? - it would be enough to do the first section's username (e.g. @ivan.dimitrovv) and public image.

A small issue with this is that the GET endpoint for users only works for the owner of the account (e.g. only the admin can GET their profile data), at least on my end (I have attached a screenshot with the rejection). Trying to fetch the avatar directly would also prove problematic, because there is no way to figure out what its file extension is.

forbidden


If there is a way to fetch the user's data directly (or alternatively, I figure out a way to fetch the user's folder and retrieve the first image), I would be happy to know

> Also, sorry if it was missed in the Jira task, but could you also do the `Public info` part of the profile? - it would be enough to do the first section's username (e.g. @ivan.dimitrovv) and public image.

julkascript:
> > Also, sorry if it was missed in the Jira task, but could you also do the `Public info` part of the profile? - it would be enough to do the first section's username (e.g. @ivan.dimitrovv) and public image.
> 
> A small issue with this is that the GET endpoint for users only works for the owner of the account (e.g. only the admin can GET their profile data), at least on my end (I have attached a screenshot with the rejection). Trying to fetch the avatar directly would also prove problematic, because there is no way to figure out what its file extension is.
> 
> forbidden
> 
> If there is a way to fetch the user's data directly (or alternatively, I figure out a way to fetch the user's folder and retrieve the first image), I would be happy to know
> 
> EDIT: I also realized that I would theoritically need a way to fetch a user by their username, since the username is used in the URL.

@milensski @dinocom33 do you understand @RyotaMitaraiWeb 's request? Can you guys pick it up?

julkascript (approved):

PR #47, opened 2024-01-21
Backand/search by user name

dinocom33:
This PR represents search user by username.

### DIFF
--- a/backend/accounts/views.py
+++ b/backend/accounts/views.py
@@ -1,14 +1,19 @@
 from django.contrib.auth import get_user_model
+from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema
-from rest_framework import viewsets, generics, serializers, status
+from rest_framework import viewsets, generics, serializers, status, permissions
 from rest_framework.decorators import action
-from rest_framework.exceptions import PermissionDenied
+from rest_framework.exceptions import PermissionDenied, ValidationError
 from rest_framework.parsers import MultiPartParser, FormParser, JSONParser
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import IsAuthenticated, IsAuthenticatedOrReadOnly
 from rest_framework.response import Response
 from rest_framework_simplejwt.views import TokenObtainPairView
 
 from accounts.serializers import RegistrationSerializer, MyTokenObtainPairSerializer, UpdateUserSerializer
+
+from accounts.filters import UserFilter
+
+from accounts.permissions import IsOwnerOfObject
 
 User = get_user_model()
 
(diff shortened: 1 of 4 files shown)


backend/accounts/views.py
```diff
@@ -38,38 +43,35 @@ def create(self, request, *args, **kwargs):
...
+    filterset_class = UserFilter
+    http_method_names = ['get', 'put', 'patch', 'delete']
 
-    def retrieve(self, request, *args, **kwargs):
-        """
-        View for update the user personal information and user delete \n
-            - API endpoint for get user information. : /api/accounts/user/<int:pk>/
-        """
-        user = self.get_object()
+    def list(self, request, *args, **kwargs):
```
julkascript, line 60: The `list` endpoint is meant to be a "get all entities" endpoint. The standard API practice is something like this:
- GET `/api/accounts/user` -> gets all users (public route, can optionally make use of query params)
- GET `/api/accounts/user/<id>` -> gets details about a specific user (public route)
- GET `/api/accounts/user/<id>/private` -> gets all sensitive details about a specific user (private endpoint - does not have to be named private in the URL, though, could be anything else, especially if it's related to some functionality)
- POST `/api/accounts/user`
- PUT, PATCH, DELETE `api/accounts/user` -> private endpoint

julkascript: So therefore, I would recommend you to restructure it like this unless you have some other suggestion, or if I missed anything?

dinocom33: Point taken. The code was refactored as follows:
 - The "list" endpoint now is public and shows user information requested by frontend.
    If there is no username as a query param the result is as bellow:
      ```
      [
        {
          "username": "simo1",
          "avatar": null
        },
        {
          "username": "simo",
          "avatar": "http://127.0.0.1:8000/media/avatars/user_simo/assassins.jpg"
        }
      ]
      ```
    If username present as a query param:
    ```
    {
      "username": "simo",
      "avatar": "http://127.0.0.1:8000/media/avatars/user_simo/assassins.jpg"
    }

 - The retrieve endpoint shows full information for registered user:
    ```
    {
      "id": 17,
      "username": "simo",
      "email": "simo@simo.com",
      "first_name": "Simo",
      "last_name": "Simov",
      "phone_number": null,
      "city": null,
      "shipping_address": "",
      "avatar": "http://127.0.0.1:8000/media/avatars/user_simo/assassins.jpg"
    }
   ```

I hope I understood the comments correctly and the changes in the code are in the right direction.

julkascript: @RyotaMitaraiWeb, are the changes sufficient? Is what you requested satisfied?

RyotaMitaraiWeb: Based on tinkering with it in Swagger, it looks perfect, great job! I will move forward with implementing it on the front end

julkascript (approved):

PR #49, opened 2024-01-24
Backend / Order and ShoppingCart implementation

milensski:
Hi,
This PR represents the shopping cart functionality and order functionality 

> [!IMPORTANT] 
**DISCLAIMER**
> I have tried following the model in Figma as it as but I hit an "dead end" and implemented it as I understand it. I really hope that this is okay with you and I will take notes if the decision is wrong.

First will start will explaining the logic behind the ShoppingCart which will lead to the Order logic as well : 

The main idea is the same: 
- The user can add items to his shopping cart at this endpoint `/api/cart/items/` with **POST** method,
- The **GET** method to the same endpoint will return all added items for the user *from the request*. 
 For this reason I have also included filtering the cart items by listing_username `/api/cart/items/?listing_username=<username>` , I will reformat or add more filters if needed. *Please note that I have made updating the quantity of the listing to be at the same POST endpoint. It just re-writes with the new value.*
- The Checkout Button - a **POST** request to the `/api/cart/items/checkout/` with *"delivery_address"* only in the body will group the cart items by sender_user and create seperate orders for each sender_user. This also triggers a request to previously implemented `mark_as_sold` method which will mark as sold and unlist a Listing if the quantity of that listing after checkout reaches 0. For this case I have disabled the public endpoint `/api/listing/<id>/buy/` but Django still can make requests to it. I am not sure for this implementation if its the best, please let me know if I need to rething the idea.

After checkout the orders can be listed with **GET** method for all orders for *all users* to `/api/order/` 
for this case I have added filters to search by *sender_user*, *receiver_user* or *status* 
also each order could be viewed separately at -> `/api/order/<id>/` 

I have included everything in the admin panel but still I could have missed something or could need changing.

For the Tests: 
I have disabled the previously written tests for `mark_as_sold` method since I have disabled the public path and the tests dont pass, but I have wrote new tests in the ShoppingCart app where I test this logic and also the whole viewset🔢 

> [!NOTE]  
> Please let me know what you think of the whole implementation and what should be changed.
> Also I will be happy to assist with more information if I have missed something in my explanation here. I tried adding everything to the Swagger as well.


julkascript:
I like your twist, I even think your idea is better and more straightforward. Nice! LGTM!

julkascript (approved):

milensski:
Awesome, I am glad that the idea works! LGTM 🚀

PR #50, opened 2024-01-27
Implement the shopping cart on the client

RyotaMitaraiWeb:
This PR makes the shopping cart functional on the client, working with the API (as far as I have tested). It is important to note that I had to do a few design changes from the screenshots I had posted in the Discord server:
- the screenshots came from a prototype that stored the data in localStorage. I did this initially so that I would have an easier time designing the page and implementing some of the features with the idea that I'd refactor them when the API became available.
- the "Back to summary" and "Go to checkout" links on mobile have been removed, since the results are now paginated.
- a context is used here so that the shopping cart button knows the amount of items (which is relevant for the badge it displays) but the state is overall maintained via the API.
- because the set code and rarities are not included in the shopping cart data (as far as I have seen), I had to cut them out of the page. My prototype used a context to pass this additional data, but seeing as the cart uses the API to maintain the state, I had to cut them out. I replaced them with hyperlinks to the card page and the seller's profiles, could probably include other useful data as compensation

also, I haven't had an opportunity to test this, but how would I calculate the total price of the checkout if the user has more than 10 listings in their cart? I am asking because the pagination prevents me from querying the user's entire cart and therefore I would only be able to calculate the total price of the current page's 10 items


julkascript:
> also, I haven't had an opportunity to test this, but how would I calculate the total price of the checkout if the user has more than 10 listings in their cart? I am asking because the pagination prevents me from querying the user's entire cart and therefore I would only be able to calculate the total price of the current page's 10 items

Good one. The total price of the cart can be provided by another endpoint. However, for the MVP purposes, we are not going to handle this use case.

julkascript:
Sorry if it was not properly defined or described, however, there are some things to adjust:
1. I don't know if it was on purpose, but I see that you have made the shopping cart flat - instead of separating all orders from different users to different sections (like in the figma), you included all of them in a single one, single table. This also works, so you might as well leave it like that. However, we need to add the rarity of the card, because this is important.
    - Since we'll have a flat table, remove the top username (that points to the logged in user's own username) and gray line
3. Bug: clicking on the card in the shopping cart leads you to a wrong card's id
4. The default behavior of the add to cart button should not be to redirect you to the cart. We should just see the red dot popping up on the cart button (and in the future - the toast message to appear, saying "Item added to shopping cart")

> However, we need to add the rarity of the card, because this is important.

RyotaMitaraiWeb:
> I don't know if it was on purpose, but I see that you have made the shopping cart flat - instead of separating all orders from different users to different sections (like in the figma), you included all of them in a single one, single table. This also works, so you might as well leave it like that. However, we need to add the rarity of the card, because this is important.

I was under the impression those different sections were just different examples, which is why I implemented it like this. In hindsight, what you are describing probably makes more sense, but what's done is done. I have removed what you requested. Once the rarity (and I guess the set code) is added to the response body, it will be reintroduced as well.

> Bug: clicking on the card in the shopping cart leads you to a wrong card's id

whoops, this should hopefully be fixed (I accidentally specified the listing ID and not the card in set ID)

> The default behavior of the add to cart button should not be to redirect you to the cart. We should just see the red dot popping up on the cart button (and in the future - the toast message to appear, saying "Item added to shopping cart")

Redirection has been removed. Toasts will be added once this PR is merged (I am currently plastering toasts around the app in a separate branch, but the toast functionality itself has been integrated, so the toast addition for the cart should hopefully come shortly after the merge)

Let me know if I missed anything

milensski:
Yes, I will try to add it tonight.
For the grouping by sender, the endpoint supports query parameter by listing_username which is actually the sender if that helps.

milensski:
@RyotaMitaraiWeb @julkascript Please check PR #51  and advise if this solves it now.

RyotaMitaraiWeb:
Rarity and set code have been reintroduced. The link to username has been removed, could reintroduce it easily

julkascript (approved):

PR #53, opened 2024-02-01
Integrate toast messages across the app

RyotaMitaraiWeb:
This PR implements toast messages. I created myself a utility function that handles toast messages for HTTP errors.

I tried to add toast messages where I felt it made sense (e.g. changing the page probably doesn't warrant a toast message). I have tested mostly the ones related to successful actions. My tests with error toast messages were limited to the login page (for a wrong username/password) and testing for irregular errors (e.g. status 500 errors) since client-side valdiation should ideally reduce the need for toast messages. This is also because I had to plaster a lot of toasts, which would make it burdensome to test manually.

It is possible that I have missed an area of the app where there could be a toast message. Similarly, I may have placed a wrong message for a particular scenario. If you spot an instance of either of those, feel free to let me know


julkascript (approved):

julkascript:
You can merge it when the conflicts are fixed

Issue #67, opened 2024-02-13
Fix the breadcrumb, its color and behavior


julkascript:
We want the Breadcrumb to be functional and a reusable component. The user should be redirected to the according page when clicking on each item of the breadcrumb.

https://mui.com/material-ui/react-breadcrumbs/

![image](https://github.com/julkascript/cardflow/assets/59143945/4621583c-deaa-4d7a-8409-91d37a56bea8)

PR #69, opened 2024-02-15
Create a breadcrumb and fix current issues related to the old one

RyotaMitaraiWeb:
This PR tackles #67 by creating a reusable breadcrumb (which utilizes the one provided by MUI). The component accepts an array of data for each link to make it more flexible and the type for the link data object is also exported so that consumers can define their breadcrumb navigation without cluttering the JSX area of the component significantly.

The breadcrumb also applies colors correctly, with the links having a gray color and the last item (the heading) having a black color.

This PR also introduces a slight change to the ``PageHeader`` component. To be more specific, if ``heading`` is a non-string ReactNode, the component will directly render that instead of rendering it in an ``h1``. The behavior remains unchanged if the heading is a string.

Disclaimer: currently, the "Buy" link items redirect to the search page, as I am not sure where it's supposed to actually lead to. There is also one page (the card details page) where there's another breadcrumb link (/ Search /) that leads to the same page. If it's supposed to link to somewhere else, let me know, as this is an easy fix. Let me know as well if I missed an area where I didn't apply the breadcrumb

### DIFF
--- a/frontend/src/components/BreadcrumbNavigation.tsx
+++ b/frontend/src/components/BreadcrumbNavigation.tsx
@@ -17,6 +17,7 @@ type BreadcrumbNavigationProps = {
 };
 
 const fontSize = '1.875rem';
+const lineHeight = '2.25rem';
 
 /*
  *  Standard separator from MUI doesn't scale alongside
@@ -38,6 +39,7 @@ function BreadcrumbNavigation(props: BreadcrumbNavigationProps): JSX.Element {
           color="inherit"
           href={l.href}
           className="flex justify-self-start"
+          lineHeight={lineHeight}
         >
           {l.text}
         </Link>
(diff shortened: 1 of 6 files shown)


julkascript:
"Buy" will redirect to the "/buy" page, which will show all the trending stuff. 

The implementation looks good. I saw something but it's up to you if you want to fix it - the font weight is not the same for every page. For example, it's different at `/sell/manage` than `/search`.

RyotaMitaraiWeb:
> "Buy" will redirect to the "/buy" page, which will show all the trending stuff.
> 
> The implementation looks good, I approve. I saw something but it's up to you if you want to fix it - the font weight is not the same for every page. For example, it's different at `/sell/manage` than `/search`.

those should be addressed now

julkascript (approved):

Issue #57, opened 2024-02-13
Create a Changelog page


julkascript:
Follow the Figma design

julkascript:
Figma design is now available [here](https://www.figma.com/file/GpsRZnN69f29pOCkiciWD4/Cardflow?type=design&node-id=1679-2&mode=design&t=yRXerYiwfQizsqjc-0)

PR #74, opened 2024-02-18
Create changelog page

RyotaMitaraiWeb:
This PR tackles #57. I am providing [screenshots of the design](https://imgur.com/a/cvk3qY1)

The content in the page is **generated** via an array in ``/src/constants/changelogArchive.ts``. I have left a JSDocs comment in there that explains how to work with it. This means that for every version release, all we have to do is simply update that array and this will be reflected in the page.

I have also made the page responsive. The current mobile design centers the Features/Fixes/Deprecated headings and aligns the bullet points to the left. In the screenshots above, I have also showcased a potential design where the headings are left-aligned as well for better symmetry, so I would appreciate feedback on which design is preferred

This PR also updates the navigation Buy/Search/Trade tabs bar to include a link to the main about page (which doesn't exist currently). With that said, I think it should receive some refactoring since it could potentially use the tabs component provided by MUI rather than use the current custom implementation and I remember even encountering cases where none of the links are selected (though I would need to investigate the steps for reproduction). It would be more appropriate to tackle this in a separate PR and I will likely open an issue regarding this at a later point and then fix it myself.

### DIFF
--- a/frontend/src/components/about/ChangelogBullet.tsx
+++ b/frontend/src/components/about/ChangelogBullet.tsx
@@ -4,7 +4,7 @@ import ErrorIcon from '@mui/icons-material/Error';
 import { SvgIconTypeMap } from '@mui/material';
 import { OverridableComponent } from '@mui/material/OverridableComponent';
 
-type changelogBullet = 'feature' | 'fix' | 'deprecation';
+export type changelogBullet = 'feature' | 'fix' | 'deprecation';
 
 type ChangelogIcon = {
   BulletIcon: OverridableComponent<SvgIconTypeMap<object, 'svg'>> & {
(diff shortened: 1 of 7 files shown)


julkascript:
1. Good job
2. You can remove the v0.9 version because it was just a filler
3. I personally think the way it is now is good. On bigger screens, we have the version to the left and the rest of the content in the middle. On smaller screens everything is in the same column, which looks good. Definitely needs to be left-aligned, though!
4. Feel free to rework the nav to be a MUI component, I also think this is a good change (in this PR or in another one)

RyotaMitaraiWeb:
Removed 0.9.0. Nav update will likely be in another PR (might be better to queue it for 1.2 since I will be focusing my attention to page implementation for 1.1), will open an issue for myself once I figure out the steps to reproduce the bug I mentioned.



> 3\. Definitely needs to be left-aligned, though!
Just to clarify, do  you mean that the headings must be aligned to the left (aka start from the same line as the bullet points) or that the container itself needs to be aligned to the left

julkascript:
> Just to clarify, do you mean that the headings must be aligned to the left (aka start from the same line as the bullet points) or that the container itself needs to be aligned to the left?

In my opinion, in mobile view, the headings aligned to the left (same line as bullet points) looks better.
IMG_0997

RyotaMitaraiWeb:
done and this should be ready for merging

PR #76, opened 2024-02-22
Added send email from contact form functionality

dinocom33:
This PR represents the backend part of contact form functionality. This functionality saves the submitted form in the database and sends an email.
Email host settings have been added to settings.py and sample.env files. To use this functionality, the real email user and password must be added to your .env file.

The url is:

http://127.0.0.1:8000/api/contacts/


julkascript:
The idea of this PR is to just get 2 fields of a form and send an email. Afterwards, we don't necessarily want to keep the data in our database, unless we want to expand this functionality in the future (which I don't think we are going to).

Currently, just to send a mail, we have created about 8 files, which don't do much and they likely won't be scaled. I personally think it's better to just have one basic function (maybe in the `accounts` app) that serves as an endpoint and view. It will send the mail without additional complexity and files. I think this way it's simple and light. But I also want to hear what you guys think? Maybe I can be missing something. @milensski @ZhivkoTringov @dinocom33

dinocom33:
I think this is the right way following good practices and SOLID principles, but if everyone thinks it doesn't need to be in a separate app I will refactor the code.

milensski:
In this case, given the simplicity of the functionality and the current requirements, the suggestion to keep it within the accounts app seems reasonable.
But in the future if the app will sent email with notifications - it is going to have a significant amount of email-related functionality, it might be a good idea to create a separate app for handling all email-related tasks. This could include the contact form emails, notification emails, and any other types of emails the application might send.

ZhivkoTringov:
I agree with @julkascript and @milensski. If we don't want to expand this funcionality in the future, I think we should keep it as simple as possible.

dinocom33:
The code for this functionality has been refactored according your comments. The URL stays the same.

milensski (approved):

ZhivkoTringov (approved):

julkascript (approved):

Issue #73, opened 2024-02-18
Fix node modules not appearing locally for development


julkascript:
When you run the containers, `npm install` is performed on the container. However, the node modules do not appear locally. This is a problem because frontend developers' extensions cannot work, unless they run a `npm install` locally. Make it so the node modules that are on the container are also appearing locally.

PR #77, opened 2024-02-23
Map the node_modules folder from container to host

milensski:
This PR is targeting issue #73 .

To achieve this task  I have used a simple solution by setting the storage location of the volume to correspond with the `node_modules` directory of the **host**. To do this, a named volume with a local driver and a bind mount has been utilized. 

Synchronizing the `node_modules` folder from the *container* to the *host* is established at creating the container with docker compose. Please note that this action is time consuming and it used around 500sec to be created.
>[!NOTE]
> My suggestions is to delete the current node_modules folder and then start this process because otherwise might not sync as intended.

Please test it and let me know if you approve this approach.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - cardflow
     volumes:
       - ./frontend:/usr/src/app
-      - /usr/src/app/node_modules
+      - node_modules:/usr/src/app/node_modules
     tty: true
     ports:
       - "5173:5173"
@@ -72,3 +72,11 @@ services:
 networks:
   cardflow:
     driver: bridge
+
+volumes:
+  node_modules:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: ./frontend/node_modules


dinocom33 (approved):

julkascript:
I receive an error at build time:
image

I have deleted the node_modules folder at the frontend and also cleaned my docker cache.

dinocom33:
I successfully build the images and ran the containers. There were no errors.

image

julkascript:
> I successfully build the images and ran the containers. There were no errors.
> 
> image

Do you see the node modules locally in `frontend/node_modules`?

dinocom33:
> > I successfully build the images and ran the containers. There were no errors.
> > image
> 
> Do you see the node modules locally in `frontend/node_modules`?

Yes. The modules are there.

julkascript:
So I was able to resolve the error, I guess I had cached stuff or something else. However, I still don't see the `node_modules` locally in my `/frontend/node_modules`.

@milensski if you delete the `node_modules` folder, and then you `docker compose up`, are the node modules appearing locally?

milensski:
> So I was able to resolve the error, I guess I had cached stuff or something else. However, I still don't see the `node_modules` locally in my `/frontend/node_modules`.
> 
> @milensski if you delete the `node_modules` folder, and then you `docker compose up`, are the node modules appearing locally?

Yes, i delete the node_modules folder, the container and the volume and after `docker compose up -d` everything is created again. I have tried 2 times today and it worked.
image

The error on the picture you shared indicates problems with absolute path, is it something with the OS setup ? 
The relative path is used to map the locally created volume with the container to the specified path so they can be placed in the right folder `frontend/`

```docker
 Line 82:     device: ./frontend/node_modules
```

Issue #56, opened 2024-02-13
Create Info functionalities


julkascript:
1. (Frontend) Information about the app
    - Who are we?
    - Quick intro into the app
    - What games we currently support on the platform (visual)
3. (Frontend) Link to the repo (for people to contribute)

-> Links to:
1. (Frontend and Backend) Contact form to reach us 
2. (Frontend) FAQ

All the designs will be in Figma.

julkascript:
Figma design [available here](https://www.figma.com/file/GpsRZnN69f29pOCkiciWD4/Cardflow?type=design&node-id=214-1003&mode=design&t=0bAGx0pyaKso4JyW-0).

PR #78, opened 2024-02-23
Implement info pages on the front end

RyotaMitaraiWeb:
This PR implements the frontend part of #56. You can preview the designs here: [https://imgur.com/a/EE3PnL1](https://imgur.com/a/EE3PnL1)

The details about each page will be separated in three sections.

**About**
The logo component has been modified with the option to change the tilda's color.

Regarding the card game images: the effects are applied with CSS and not with image-editing software. The images are saved in an ``assets`` folder that was also created in this PR. The red "coming soon" text may overlap with red areas of the image; the only solutions to this are either hardcode the position for each image or to change the color. I am not a huge fan of the former as it is hard to control for different viewports. My suggestion is to keep them as they are, as the effects will eventually be removed as we support more games. **Note:** the images have varying heights, so the "coming soon" texts will not be equally aligned (as they are placed near the bottom of each image)

There are two minor design tweaks that I have gone with:
- the "Buy, sell and trade for free" and "Join the community" texts have been turned into headings. The original design didn't look good to me, mainly due to the different colors. Semantically speaking, those make more sense as headings anyways. This also leads to a more consistent design, as that section would be the only one without headings otherwisse.
- Some icons may be different from the wireframes due to me not finding the equivalent in MUI's icon library, in which case I have gone with whatever felt the closest.

**FAQ**
The FAQ page, like the changelog page, is generated from a ``constants`` file. The answers can be provided as any valid React node; this is to allow more complex markup like hyperlinks and formatting. I have tried to provide answers to the questions based on my own understanding of the project, though I did leave one placeholder answer due to me having absolutely no idea what to put in there. If there are errors or desired changes, feel free to let me know or modify the FAQ data yourself.

**Contact us**
The form sends a request to the contacts API endpoint upon submission. As far as I have tested, the request itself is sent properly, but I have no idea whether an email would be sent correctly, as I have not had an opportuntiy to set up an email for local development (interestingly, I got a 200 OK response despite using the data from the sample .env file). The fields are validated to prevent unnecessary calls to the server. The email field validation is also debounced as to not hit the user with errors as soon as they type.

===

Let me know if there are any errors or other desired changes.


julkascript:
Awesome, it's literally a copy of the design! 👏

I have just one idea to improve the look of the more info bar. Maybe we can align the items to the left with a bit of a gap. I think this way it looks slick:
image

julkascript (approved):

PR #80, opened 2024-02-26
Map the node_modules folder from container to host 2

julkascript:
This version of #77 works for me. It may be because I am working under WSL2 (Ubuntu). 

@milensski @dinocom33 Can you see if this also works for you guys? Also, I accept feedback if you have something specific in mind or ideas for improvement.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,11 +15,10 @@ services:
       - cardflow
     volumes:
       - ./frontend:/usr/src/app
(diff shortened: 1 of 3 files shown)


dinocom33:
This version works for me as expected as the previous one. All necessary files  and folder are in place and the app is working.

dinocom33 (approved):

milensski (approved):

PR #82, opened 2024-02-28
Backend/future#62 feedback

dinocom33:
This PR represents the backend part of rating and feedback functionality.
A new method has been added to the user serializer - ```get_average_rating```, that calculates the average rating for the user.

To view feedbacks and ratings for all users use the following endpoint:

```
/api/feedback/
```

the response:
```

[
  {
    "rated_user": 3,
    "order": 1,
    "given_from": 4,
    "user_rating": 5,
    "comment": "very good"
  },
  {
    "rated_user": 2,
    "order": 2,
    "given_from": 4,
    "user_rating": 5,
    "comment": "not so good"
  },
  {
    "rated_user": 2,
    "order": 3,
    "given_from": 4,
    "user_rating": 1,
    "comment": "bad"
  },
  {
    "rated_user": 2,
    "order": 4,
    "given_from": 4,
    "user_rating": 3,
    "comment": "good"
  },
  {
    "rated_user": 2,
    "order": 6,
    "given_from": 4,
    "user_rating": 1,
    "comment": "so bad"
  },
  {
    "rated_user": 2,
    "order": 7,
    "given_from": 3,
    "user_rating": 3,
    "comment": "so so"
  },
  {
    "rated_user": 4,
    "order": 8,
    "given_from": 2,
    "user_rating": 3,
    "comment": "not so bad"
  }
]
```

To view the average rating and all comments for specific user use the following endpoint:
```
/api/feedback/user/{id}/
```

the response:

```
[
  {
    "user": 3,
    "average_rating": 4,
    "all_comments_and_ratings": [
      {
        "related_order": 2,
        "given_from": 4,
        "rating": 5,
        "comment": "very good"
      },
      {
        "related_order": 11,
        "given_from": 2,
        "rating": 1,
        "comment": "string"
      },
      {
        "related_order": 12,
        "given_from": 2,
        "rating": 4,
        "comment": "string1"
      },
      {
        "related_order": 13,
        "given_from": 2,
        "rating": 5,
        "comment": "string11"
      },
      {
        "related_order": 17,
        "given_from": 4,
        "rating": 5,
        "comment": "very good seller"
      }
    ]
  }
]
```


julkascript:
Overall, great job. I just think that the `given_to` and `given_from` should be the same as the `sender_user` and `receiver_user`, derived from the order. I don't think there should be a choice here, because this opens for mistakes. @milensski what do you think?

milensski:
Yes, overall looks good. 
I was wondering the same for these variables but was not sure if its just me. So I agree with that.

dinocom33:
I thought the same thing when I started, but I thought it might get confusing with the variables, so I'll change them. Regarding where the rating and comments are, what do you guys think, isn't it more correct to have it in the account app, since both are about users?

backend/order/models.py
```diff
@@ -43,3 +44,20 @@ class OrderItem(models.Model):
...
         return f'{self.listing} : {self.quantity}'
+
+
+class FeedbackAndRating(models.Model):
+    given_to = models.ForeignKey(User, on_delete=models.CASCADE, related_name='given_to')
+    given_from = models.ForeignKey(User, on_delete=models.CASCADE, related_name='given_from')
+    related_order = models.ForeignKey(Order, on_delete=models.CASCADE, related_name='related_order')
+    rating = models.PositiveIntegerField(
+        default=1,
+        choices=[(i, i) for i in range(1, 6)],
```
milensski, line 55: @julkascript If you mean this line with choices I think the expression should not cause any error since in the migration it is transfromed to numbers.

backend/order/migrations/0001_initial.py
```diff
@@ -34,4 +35,15 @@ class Migration(migrations.Migration):
                 ('order', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='order.order')),
             ],
         ),
+        migrations.CreateModel(
+            name='FeedbackAndRating',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('rating', models.PositiveIntegerField(blank=True, choices=[(1, 1), (2, 2), (3, 3), (4, 4), (5, 5)], default=1, null=True, validators=[django.core.validators.MaxValueValidator(5)])),
```
milensski, line 42: @julkascript here the choices are transformed to array of tuples with just number.

dinocom33:
The variable names have been changed according to the comments.

julkascript:
No, guys, maybe I expressed myself wrong, but I meant something completely different.

What I wanted to say was that when a feedback object is created, we should be taking the `receiver` and `sender` users from the order itself programmatically. Because what can currently happen is that you can have a completed order, where `buyer_1` bought from `seller_1`. But in the backend, it can happen that `buyer_6` can give the feedback to `seller_1`, even if that buyer has never been part of that order. Currently in the API, I don't see anything stopping me from doing that (correct me if I am wrong).

There are 2 ways I know on approaching the problem:
- Fix the issue on the data side - ensure that the feedback will always use the `sender_user` and `receiver_user` from the order. 
- Fix the issue on the controller side - ensure that through the views and serializers you match both users correctly and all use cases are covered. I would personally go for option 1, but in both cases, the controller side needs to be good.

What do you think? Do you get what I mean?

dinocom33:
Sorry for misunderstanding. I thing now it is as expected.

julkascript (approved):

Issue #65, opened 2024-02-13
Save card pictures in the backend


julkascript:
Save the pictures of the cards locally and serve as static images. According to [the API](https://ygoprodeck.com/api-guide/), if we do not do this, we may get banned by IP from using it.

julkascript:
> @julkascript hi there, i would like to work on this, please assign this to me Thanks

Hey. Feel free to do so. But do you have the context? Do you need assistance?

HyoungSooo:
Sure! thanks to @julkascript.
what should i do?

julkascript:
> Sure! thanks to @julkascript. what should i do?

- Check out the [repository README](https://github.com/julkascript/cardflow/blob/develop/README.md) to see how to set up the project locally for development. 
- Also, check out the [contributing guidelines](https://github.com/julkascript/cardflow/blob/develop/CONTRIBUTING.md) to have some idea on how we contribute around here! 

Happy coding! Let me know if you need more help.

PR #85, opened 2024-03-21
Show image on camera hover during search

julkascript:
It's written in the issue that it should be blocked by #65 but it's actually not because sorting out what image will go to the backend is a separate issue. We will always only show one here. Lmk if I missed something.

image


RyotaMitaraiWeb (approved):
No issues on my end, good job

Issue #61, opened 2024-02-13
Finish the purchase functionality


julkascript:
1. (Frontend and Backend) Sellers should receive the sell request when a user buys their listing.
- "My sales" page, where the seller:
  - receives sell requests
  - can mark them as sent or rejected
  - Note: When the sell request has been marked as sent, the status of the order changes to "sent"
  - Note: When rejected - restore the numbers (status of order changes to "rejected")

2. (Frontend and Backend) Buyers should be able to track their purchase: see the state of the purchase (sent or not) and confirm if it's received
- Buyers have a "My orders" page, where they:
  - see their past orders listed, and their state
  - they can confirm an order that is in status "sent" -> order becomes "completed"
  - they can reject an order that is in status "sent" -> order becomes "rejected"
  - seller receives in his sell request list that the order is completed on the buyer's end
  - if the order has not been completed in 7 days, it becomes auto-completed

The `My orders` and `My sales` pages shall be accessed from the user dropdown, which is  implemented in #68.
Follow the Figma designs.

julkascript:
Designs:

- [My sales](https://www.figma.com/file/GpsRZnN69f29pOCkiciWD4/Cardflow?type=design&node-id=1712-461&mode=design&t=gqWrWHsoq24z0YfN-0)
- [My sales modal](https://www.figma.com/file/GpsRZnN69f29pOCkiciWD4/Cardflow?type=design&node-id=1712-535&mode=design&t=gqWrWHsoq24z0YfN-0)

julkascript:
@milensski @RyotaMitaraiWeb Guys, any movement here? Any difficulties?

RyotaMitaraiWeb:
I have most of the UI ready and am just waiting for the backend so I can make it functional

milensski:
I am done with the functionality, only the tests are missing and I am struggling with the auto complete job - not triggering

julkascript:
> I am done with the functionality, only the tests are missing and I am struggling with the auto complete cron job - not triggering as expected

@milensski For me, it's okay if you already put the functionality on a PR. The "auto-complete" job we can do another time - it's not that important.

milensski:
@julkascript @RyotaMitaraiWeb 
Perfect, I was also able to finish my tests today and created the PR.

Let me know what are your thoughts.

PR #86, opened 2024-04-12
Backend / Purchase functionality

milensski:
Hi, 
This PR is implementation for the issue #61.

Most of the functionality was already implemented in PR #49   so the new part is :

PATCH Request for updating the status of the order will update the order if the user is the sender or receiver and also:
 - this creates Order History record in a new table to track order statuses 
 >[!IMPORTANT]
 > Make sure to rebuild the image in docker when testing locally

 - IF the order is changed to `rejected` it restores the purchased quantity to the listing as well as the *is_sold* and *is_listed* flags and it will lock it for further updates.
 - If the order is `completed` it will also lock the status update.

The custom command for auto-completing all orders on status sent or above more than 7 days has also been created, but unfortunately, I was not able to integrate it to work in the docker container with cronjob (the cronjob is created but for some reason it is not triggered)
It is in the changes of this PR and does not interfere with the current flow of the server, it can be debugged later. 
Manually executing the command is working. `python manage.py complete_old_orders` will trigger it.

I have added 3 tests as well.

Please review and if something is missing, I will add it or change it. Thank you!

> I want to apologize  for my absence and for the caused delay in the project, I am now here and able to contribute again!


backend/entrypoint.sh
```diff
@@ -31,19 +33,27 @@ done
...
 
-if [ "$(python manage.py shell -c 'from accounts.models import User; print(User.objects.filter(username="admin", email="'"$PGADMIN_DEFAULT_EMAIL"'").exists())')" = "False" ]; then
+exists=$(python manage.py shell -c 'import django; django.setup(); from django.contrib.auth import get_user_model; User = get_user_model(); print(User.objects.filter(username="admin").exists())' | tr -d '\n')
+
+if [ "$exists" = "False" ]; then
   echo "Creating superuser..."
-  python manage.py shell -c 'from accounts.models import User; User.objects.create_superuser("admin", "'"$PGADMIN_DEFAULT_EMAIL"'", password="'"$PGADMIN_DEFAULT_PASSWORD"'")'
-  echo "Superuser is created"
+  python manage.py shell -c 'import django; django.setup(); from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser("admin", $"PGADMIN_DEFAULT_EMAIL", password=$"PGADMIN_DEFAULT_PASSWORD")'
+  echo "Superuser created"
```
julkascript, line 44: This is not a valid bash syntax. The quotes need to surround the variable name and the $. I changed it to this and it works:

```
python manage.py shell -c "import django; django.setup(); from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', '$PGADMIN_DEFAULT_EMAIL', password='$PGADMIN_DEFAULT_PASSWORD')"
```

julkascript:
Are the following migrations applied in production? I think not, but can you confirm, please?
```
0002_rename_given_to_feedbackandrating_receiver_user_and_more.py
0003_alter_feedbackandrating_receiver_user_and_more.py
```

- If they are not applied, you can totally delete them, no problem - even better, keep the number of migrations low.
- If they are applied, we will have to drop the entire db.

milensski:
> Are the following migrations applied in production? I think not, but can you confirm, please?
> 
> ```
> 0002_rename_given_to_feedbackandrating_receiver_user_and_more.py
> 0003_alter_feedbackandrating_receiver_user_and_more.py
> ```
> 
> * If they are not applied, you can totally delete them, no problem - even better, keep the number of migrations low. 1 migration per version is perfect.
> * If they are applied, we will have to drop the entire db.

As far as I know there are not applied, @ZhivkoTringov  or @dinocom33 , can advise If I am right.

backend/entrypoint.sh
milensski: Yes, you are correct I have missed this, but I want to double check why it is working or not in my docker.

dinocom33:
> > Are the following migrations applied in production? I think not, but can you confirm, please?
> > ```
> > 0002_rename_given_to_feedbackandrating_receiver_user_and_more.py
> > 0003_alter_feedbackandrating_receiver_user_and_more.py
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * If they are not applied, you can totally delete them, no problem - even better, keep the number of migrations low. 1 migration per version is perfect.
> > * If they are applied, we will have to drop the entire db.
> 
> As far as I know there are not applied, @ZhivkoTringov or @dinocom33 , can advise If I am right.

They are not applied.

RyotaMitaraiWeb:
Most of the status flow works (at least as far as I understand it), but there are a few behaviors that I don't think are intended:
- the purchaser can directly mark an order as rejected or completed, even if the order has not been sent. I would think that there would be checks to make sure that you can complete or reject an order only when it's marked as sent (as far as I understand the status flow, at least)
- this is moreso a question, as I am not sure what the intended behavior is, but are *both* the seller and purchaser supposed to be able to mark the order as complete or rejected? I would think only the purchaser should be able to do that, but the seller can also mark them as such
- theoritically, you can regress the status (except when it's rejected or completed, where it correctly locks the status); one can mark the order as sent and then mark it as ordered. If this is intended, ignore this.

Also, the order API does not provide the purchased cards at all, which would prevent me from displaying them in the modal (according to the Swagger schema, there's supposed to be an ``order_items`` property, but it's missing on my end). Not sure if this is in the scope of this PR or an issue on my end, but figured I should mention this.

milensski:
@RyotaMitaraiWeb notes taken, please check now and let me know what you think, I will correct or add something if needed
 - OrderItems are now being correctly populated 
 
 > this is moreso a question, as I am not sure what the intended behavior is, but are both the seller and purchaser supposed to be able to mark the order as complete or rejected? I would think only the purchaser should be able to do that, but the seller can also mark them as such
 - This is intended behavior as described in the issue, but I also think it is more convenient that way because for example if the card is sent to the **buyer** but he return it or don't take it the **seller** should be able to reject, so the listing is restored. Or simply seller don't want to send it he should be able to reject the order.
 
 >theoritically, you can regress the status (except when it's rejected or completed, where it correctly locks the status); one can mark the order as sent and then mark it as ordered. If this is intended, ignore this.
  - I am not sure if this was required but I have made validation for not being able to downgrade the status :)
  
 @julkascript I have corrected the bash syntaxis, I have missed it because I have already a superuser in my DB. 
 Please let me know your comments.
 Thanks.

RyotaMitaraiWeb:
My concerns seem to have been addressed (looking at the Figma wireframes, it looks like outright rejection is possible, so good thing that only completion has been restricted to after being sent).

With that said, I found another issue that is hopefully an easy fix. Currently, the backend treats "received" and "completed" as different statuses, but according to the Figma wireframes, they are actually the same. As a matter of fact, if you mark an order as received, you will be unable to mark it as completed, because the backend will not allow you to mark it as sent and thus will not allow you to complete it either. Funnily enough, you can actually reject the order at this point. For consistency, we should probably stick to only one of the statuses (I would prefer "received" as it matches what is shown on the frontend)

milensski:
> My concerns seem to have been addressed (looking at the Figma wireframes, it looks like outright rejection is possible, so good thing that only completion has been restricted to after being sent).
> 
> With that said, I found another issue that is hopefully an easy fix. Currently, the backend treats "received" and "completed" as different statuses, but according to the Figma wireframes, they are actually the same. As a matter of fact, if you mark an order as received, you will be unable to mark it as completed, because the backend will not allow you to mark it as sent and thus will not allow you to complete it either. Funnily enough, you can actually reject the order at this point. For consistency, we should probably stick to only one of the statuses (I would prefer "received" as it matches what is shown on the frontend)

Totally agree with you, while checking the figma my thoughts were the same and funny enough I chose `completed` status as more intuitive for me. It will be more easy if we consider completed for absolute, but its not hard to rewrite for received.

julkascript:
> > My concerns seem to have been addressed (looking at the Figma wireframes, it looks like outright rejection is possible, so good thing that only completion has been restricted to after being sent).
> > With that said, I found another issue that is hopefully an easy fix. Currently, the backend treats "received" and "completed" as different statuses, but according to the Figma wireframes, they are actually the same. As a matter of fact, if you mark an order as received, you will be unable to mark it as completed, because the backend will not allow you to mark it as sent and thus will not allow you to complete it either. Funnily enough, you can actually reject the order at this point. For consistency, we should probably stick to only one of the statuses (I would prefer "received" as it matches what is shown on the frontend)
> 
> Totally agree with you, while checking the figma my thoughts were the same and funny enough I chose `completed` status as more intuitive for me. It will be more easy if we consider completed for absolute, but its not hard to rewrite for received.

Your remarks make sense, guys. To me, both options would be equally as fine. Maybe you can align between the both of you and come to a conclusion.

RyotaMitaraiWeb:
whether it's "completed" or "received" isn't a hill I am willing to die on, I am perfectly fine with whichever is selected. I suggested "received" purely because I felt it made more sense to match it with the frontend.

The main point is that the backend should probably use only one of them, since having both will make things more complex. Sticking with "completed" is perfectly fine, seeing as the validations for it have already been written

milensski:
@julkascript is there anything else to change in the code?
SInce we have agreed to work with `completed`, the logic in the serializer for updating is validating by status completed and there is nothing to change.

Please review and have your comments if I missed something or I need to add.

milensski:
> No more comments. I was waiting for the removal of the `received` value from the status.

I have deleted the unused statuses and made a migration since they are in the deployed version.

RyotaMitaraiWeb:
I think the status flow now works appropriately (or, at the very least, I have not found a way to break it).

The only concern that I have at this point is that the PATCH endpoint can theoritically edit the delivery address. I could maybe see this as a feature, especially if the upcoming modal allows the buyer to change the address, but I will need a clarification on this matter to be sure.

If this is not a concern, then I will approve the PR, since I cannot think of issues with the API at the moment.

julkascript:
> I think the status flow now works appropriately (or, at the very least, I have not found a way to break it).
> 
> The only concern that I have at this point is that the PATCH endpoint can theoritically edit the delivery address. I could maybe see this as a feature, especially if the upcoming modal allows the buyer to change the address, but I will need a clarification on this matter to be sure.
> 
> If this is not a concern, then I will approve the PR, since I cannot think of issues with the API at the moment.

Yes, correct observation. The delivery address shall not be able to be modified from this endpoint (as well as the receiver user).

milensski:
> > I think the status flow now works appropriately (or, at the very least, I have not found a way to break it).
> > The only concern that I have at this point is that the PATCH endpoint can theoritically edit the delivery address. I could maybe see this as a feature, especially if the upcoming modal allows the buyer to change the address, but I will need a clarification on this matter to be sure.
> > If this is not a concern, then I will approve the PR, since I cannot think of issues with the API at the moment.
> 
> Yes, correct observation. The delivery address shall not be able to be modified from this endpoint (as well as the receiver user).

Your comments have been addressed, and I have edited the update method for this endpoint restricting any updates different than the status. I have not thought about it but you have a point. The order should not be able to be changed once it is created. This means if its not correctly placed it should be rejected and then re-ordered.

@julkascript @RyotaMitaraiWeb  Please let me know if you have other concerns as well.

RyotaMitaraiWeb:
It doesn't seem like ``not received`` is implemented as a valid status

(speaking of which, does ``not received`` terminate the order like ``completed`` and ``rejected`` do? Also, is the buyer capable of marking an order as ``not received`` before it's ``sent``? This is moreso a question for @julkascript since I am currently integrating the solution on my end as well)

julkascript:
Hey. When we were discussing the statuses, we thought of “not received” as a status where something went wrong with the package. However, we can skip this status for now - we can implement later.

RyotaMitaraiWeb (approved):
I have nothing more to say then, since my bajillion concerns have been addressed by now

julkascript (approved):

dinocom33 (approved):
Looks good to me.

Issue #79, opened 2024-02-23
Issues with and enhancements for navigation menus and bars


RyotaMitaraiWeb:
There are a few enhancements that can be applied to the navigation menus and bars, as well as some bugs that need to be fixed.

**ListingTopBar**
Refactor the component to use [MUI's Tabs component](https://mui.com/material-ui/react-tabs/). You should also probably rename the component to something else, as the component is also shown in the About pages.

There is also a **bug** with the current implementation. When you search for a card via the search field and click "See all results", none of the tabs are selected.

**ProfilePictureAvatar and mobile links**
While developing the profile picture avatar navigation and its mobile equivalent, I forgot to set unique keys for each link (as they are created with loops). Make sure that the desktop and mobile keys do not collide with each other to prevent render errors.

I have slotted this for v1.2 as I want to focus on clearing up the v1.1 backlog. However, **feel free to take this up earlier if you want to work on it**. I am also willing to fix this myself if no one else is interested.

PR #88, opened 2024-04-22
Fix issues with the tabs and navigation menus

RyotaMitaraiWeb:
Addresses #79.

One important thing to note is that the tabs' underline does not have a transition, I suspect this is caused by the component getting destroyed and then mounted again when navigating to a different tab page. Perhaps a global context could be used to bypass this, but I feel like it will be an overkill for little gain (and, technically speaking, this is consistent with the previous variant's behavior).

There may be a few visual changes due to the tabs' default styling, but those should be minor.

I also renamed the component to ``CardflowTabs`` to be more generic, could maybe change it to something else.


julkascript (approved):

julkascript:
The transition is not a problem. We can fix it at a later point if it becomes necessary.

PR #89, opened 2024-04-25
Implement feedbacks on the frontend

RyotaMitaraiWeb:
This PR makes the feedback part of the orders modal functional.

For now, the frontend retrieves the feedback for each of the displayed orders in the table, which is mostly so that they can be loaded instantly when the modal is opened. Buyers can leave a review at any point of the order (which is at least what happens on the backend, since you can leave a review even in an ordered state). Whether the save button is disabled has also been updated to account for the possibility of leaving a review.

Tests have been updated and some new ones have been added to reflect the new features.

### DIFF
--- a/frontend/src/components/orders/OrdersTable.tsx
+++ b/frontend/src/components/orders/OrdersTable.tsx
@@ -1,3 +1,4 @@
+import { UserFeedback } from '../../services/feedback/types';
 import { Order } from '../../services/orders/types';
 import MarketTable from '../marketTable/MarketTable';
 import OrdersTableRow from './OrdersTableRow';
@@ -12,6 +13,7 @@ type OrdersTableProps = {
   onChangePage: (page: number) => void;
   count: number;
   updateOrders: () => void;
+  feedbacks: UserFeedback | undefined;
 };
 
 function OrdersTable(props: OrdersTableProps): JSX.Element {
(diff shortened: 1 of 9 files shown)


julkascript:
Looks and feels good, however, I found the following bug:

0. I have 1 review already in the table
1. Review order
2. Write the order and the stars
3. Save
4. Open the modal -> the review is there
5. Refresh the page, open the modal -> the review is **gone**
6. Cannot write and save a new one

Also, I've been thinking - the "Review Order" button would look nicer if it's of variant `outlined`, and color `primary`.

julkascript:
Also, let's make it, so reviews can only be left after the order is completed or rejected. I'll fix it in the backend.

RyotaMitaraiWeb:
To be more specific about the bug, I tried creating multitutes of orders and reviews from three different accounts and everything works correctly on my side, even after refreshing on all ends

RyotaMitaraiWeb:
Update: I managed to reproduce the bug on my end, so I can confirm it exists, I will investigate more

julkascript:
@milensski @dinocom33 can you guys check this bug out?

milensski:
Hi,

I was able to reproduce the bug even with 18 orders, when I made a review to order 14 it went in order 1, 
I will examine it and will fix. 

@RyotaMitaraiWeb  Thanks for looking into it , I will check your advise first.

PR #90, opened 2024-04-27
Fixing bug in the Rating and Feedback retrieving method

milensski:
As mentioned in #89 , the bug for retrieving the correct feedback and rating should now be fixed with this PR

The issue was that the zip method was returning wrongly matched data as @RyotaMitaraiWeb mentioned.

I have refactored this approach. 

Hope its solved now.


RyotaMitaraiWeb (approved):
Feedbacks are now mapped to the correct orders on the front end, great job

PR #91, opened 2024-04-28
Implement a mobile search field

RyotaMitaraiWeb:
This PR is part of my efforts to implement some of the mobile Figma wireframes. This PR focuses on the search field. [Review the UI](https://imgur.com/a/JEqcKxU)

There are some major refactors / design changes:
- the search results retrieval has been moved to its own hook so that it can be reused, seeing as we use the logic more than once. I have only applied it to the search bars in the navigation menu, but when the time comes, I will implement it elsewhere as well. I have written some basic tests for it as well.
- The shopping cart button is no longer rendered initially on the mobile UI. It is now part of the navigation menu and, similarly to its desktop counterpart, gets some highlighting when there are items in it
- I deleted a bunch of legacy code that existed in previous variants of the search field, this does not break anything as far as I am concerned.

The mobile search field consists of components separate from the desktop one's, as trying to maintain a single component that reflects both is very hard without breaking the desktop variant's design or writing very messy code (since the desktop search field overrides some of MUI's default stylings to achieve its design, which makes it harder to write responsive code). Also, the mobile search is implemented as a dialog, meaning that there is a backdrop when opened. I have opted for this approach because it better redirects the user's focus to the search field and makes it easier to locate the field itself.

**Note:** for the search results, the user may have to scroll slightly to the right to see the full name if the screen is particularly small (something like ~330px width) and the name is particularly long. I haven't addressed it, as it's not an issue on even modestly bigger resolutions and the overflow is minor at best anyways.


julkascript:
Looks and feels good. However, on devices with width less than 650, the search tends to go way too much to the middle. I think some mobile users may not have such a solid experience.
image
image

RyotaMitaraiWeb:
The issue comes from the fact that the home page is overflowing in that viewport. If you test the field on a page that doesn't overflow (like the Login page), the button is rendered correctly.

My next PR will address this (+ a few other pages)

julkascript:
oh yes, correct, nvm

julkascript (approved):

PR #92, opened 2024-04-30
Refactor new / edit listing pages and make them responsive

RyotaMitaraiWeb:
This PR is part of my efforts to implement some of the mobile Figma wireframes. This PR focuses on the listing pages. [Review the UI](https://imgur.com/a/j3qfdFB).

However, I have to note that there were some major rewrites that I had to do in order to achieve the current design, as there were a lot of things that were hardcoded. In addition, it was hard to follow what exactly was happening, so this rewrite is meant to make the code a bit more readable. Finally, there were a few bugs that outright interferred with my development, so I had to fix those to make my life easier.

===

**The rewrite**
The rewrite includes the following:
- extract the listing form into its own component for reuse
- usage of MUI components in place of custom ones (where appropriate)
- rewrite of some mechanisms to reduce unnecessary code
- a few bugfixes, which include:
- - some redirects when loading the pages
- - retrieval of resources by wrong IDs

As a result of the rewrite, there may be a few visual differences from the previous version due to MUI's default styling, though I think the main visual aspects have been preserved

**The mobile UI**
This PR includes implementations of the edit listing and new listing pages. I have tried to match the Figma wireframes as much as possible. With that said, though, a few concessions had to be made. In particular, the card image doesn't look particularly good on very small screens. We can consider some different design if it's necessary, as the current one doesn't downscale perfectly without making certain elements too small.

The new listing page that holds the search field was made the way it is because, on smaller screens, the search results do not fit in the remainder of the container and the text starts overflowing.

I have hidden the necessary table cells in accordance to the wireframes.

**Out of scope for the PR**
The following tasks have been left out of this PR, despite appearing in the wireframes and will be addressed another time. Those are:
- Breadcrumb navigation restructuring - this is something that affects the entire app, so I want to tackle this separately
- Combining seller, condition, and language cells into one - this will require some refactoring in multiple places, as, currently, all applicable cases are implemented as separate cells. There is also the question of whether that's a good idea in first place, as semantically they represent very different data


julkascript:
LGTM! And don't be afraid to change stuff or go out of the design, ya know. :) It's just a "direction". We change when we feel we want to change, and I always vouch for that. It's always a collective vision and effort.

julkascript (approved):

Issue #97, opened 2024-05-14
Add text to describe that there are currently no listings for the given card


julkascript:
Add: "There are currently no listings for this card. You can be the first one!"

![image](https://github.com/julkascript/cardflow/assets/59143945/381d5d4d-b44b-4742-8dbf-f0166940c356)

PR #103, opened 2024-05-20
Add a special text for when there are no listings

RyotaMitaraiWeb:
Tackles #97.

The pagination widget is absent when there are no listings, since it's pointless in that context

### DIFF
--- a/frontend/src/components/yugioh/table/market/YugiohCardMarket.tsx
+++ b/frontend/src/components/yugioh/table/market/YugiohCardMarket.tsx
@@ -10,6 +10,10 @@ type YugiohCardMarketProps = {
 };
 
 function YugiohCardMarket(props: YugiohCardMarketProps): JSX.Element {
+  if (props.listings.length === 0) {
+    return <TableWithNoData />;
+  }
+
   return (
     <>
       <div className="hidden lg:flex flex-col md:items-center justify-center overflow-auto">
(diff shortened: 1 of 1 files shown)


frontend/src/components/yugioh/table/market/YugiohCardMarket.tsx
```diff
@@ -60,4 +64,49 @@ function YugiohCardMarket(props: YugiohCardMarketProps): JSX.Element {
...
+          <tfoot>
+            <tr>
+              <td colSpan={9}>
+                There are currently no listings for this card. You can be the first one!
+              </td>
+            </tr>
+          </tfoot>
+        </MarketTable>
+      </div>
+      <div className="flex lg:hidden flex-col mx-auto overflow-auto">
```
julkascript, line 89: Isn't this whole div with its content a copy of the one above? Does it do anything? I also see this for the return of `YugiohCardMarket`.

RyotaMitaraiWeb: the duplicates are there because some of the cells have a different ``colSpan`` depending on the viewport (which cannot be controlled with CSS classes alone). This necessity comes from the fact that some cells are hidden on smaller viewports and this would affect the table layout significantly, so this solution came into existence.

RyotaMitaraiWeb: As for this specific case, there's also the fact that the pagination is controlled with props and the pagination is placed in a ``tfoot``, so there would be no way to insert another cell next to the pagination (+ there's no point in having the widget if there's nothing to paginate)

julkascript (approved):

Issue #95, opened 2024-05-14
Username already exists upon registration - does not show error or any response to user


julkascript:
Steps to reproduce:
1. Sign up
2. Sign out
3. Sign up again with the same username
4. See that no error or feedback message is shown (err 400 in the network tab is correct, though, the message can be taken from there).

RyotaMitaraiWeb:
there are a few other missing validations as well, I will combine them with this one

PR #108, opened 2024-06-01
Display an error if the username already exists and use client-side validation where possible

RyotaMitaraiWeb:
This PR tackles #95, with two caveats:
- as I mentioned in the linked issue, only some of the validations were performed on the client, the rest relied on the server
- I mentioned in #93 that I wanted at some point to improve the way validation was done on those pages for better UX and decided to do it now.

The reason for this is that there is no point in sending an invalid registration to the server. The validations also apply in the login page for similar reasons. The submit button is disabled when any of the fields are invalid or when at least one field has not been changed at all (inputting anything in a field will cause a debounced status update, which makes the field "dirty").

The only validation that is not performed client-side is for whether the username is already in use, as there is no way to check that before sending the request (in the case of the register page).

If the username is already in use, the error is placed in the username error, though a simple input change will clear it off.

**Note 1:** the email pattern used on the front end likely doesn't completely match the one that comes straight out of Django. The only reason I haven't tried to make them consistent is because it seems like Django performs multiple regex tests on the email and this would require some tweaking on my end. Whether this is worth doing is a different question (since all of the common email domains I could think of are valid in both cases), but in case it is, consider my current implementation more of a partial POC at this stage.

**Note 2:** this PR features a small change to the debounce hook, which makes it possible to pass arguments to the generated function. The reason for this is that I wanted to use one hook for multiple status changes (in regards to whether a particular field has been changed or not) out of laziness, but this ended up causing issues in the cases where a user fills multiple fields very quickly (in which case it did not update all fields' status), so I had to abandon this idea. So basically, there's a small change in the debounce hook that ended up totally irrelevant to the issue at hand.

**Note 3:** I haven't applied any of this in the profile page, since I want to make sure that my approach is fine before I touch that page.

### DIFF
--- a/frontend/src/components/auth/AuthForm.tsx
+++ b/frontend/src/components/auth/AuthForm.tsx
@@ -105,8 +105,6 @@ const SignUpPage: React.FC<AuthFormProps> = ({ isLogin }) => {
       const payload = userService.extractUserFromToken(tokens.access);
       await authenticaticateUser(payload.user_id, tokens);
     } catch (error: any) {
-      // const data = await error.err.json();
-      // setError(data);
       if (error instanceof HttpError) {
         if (isLogin && error.err.status === 401) {
           errorToast(error, toastMessages.error.failedLogin);
--- a/frontend/src/components/auth/AuthForm.tsx
+++ b/frontend/src/components/auth/AuthForm.tsx
@@ -13,7 +13,7 @@ import toast from 'react-hot-toast';
 import { toastMessages } from '../../constants/toast';
 import { errorToast } from '../../util/errorToast';
 import { HttpError } from '../../util/HttpError';
-import { Button, Link, TextField, Typography } from '@mui/material';
+import { Button, Link, TextField } from '@mui/material';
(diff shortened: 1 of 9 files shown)


julkascript:
It looks very good, feels nice, good job as always!

I noticed the case where you bypass the frontend validation with a wrong email and you get 400 from the backend but it's not displayed as error message. It would be a good idea to handle this. We don't need to match Django's validations but it would be enough just to display the toast message with the reason for the failure. This would apply also if you manually make the backend return other responses like 4XX or 5XX to simulate the service being down. Always a good idea to keep the user informed.

image

RyotaMitaraiWeb:
I have made it so that, if the server returns an error for the email, it will be routed to the email field. In addition, if the error is of a status code different from 400, an error toast will be displayed.

(Also, this is something I forgot to note in my initial comment, but if a field fails multiple validations, the field will only display the first error, this is fully intentional)

julkascript (approved):

julkascript:
LGTM

Issue #96, opened 2024-05-14
Fix "All-time best sellers" requirements


julkascript:
Ensure that a card can only appear once on the All-time best sellers. So "Dark Hole" as a card can only be seen there once.

PR #109, opened 2024-06-02
Fix all time best sellers cards

milensski:
This PR fixes issue #96 

Each card should be returned only once now regardless of the many listings.


dinocom33 (approved):
Looks good to me.

julkascript (approved):

PR #110, opened 2024-06-07
Download and Save images locally

milensski:
This PR fixes issue #65 

I have implemented a manage command which is similar to the seed command for downloading one, staple, or all card images.
This was for analytic purposes, but it's good to have it there, just in case.

The logic for saving the cards is simple and if it's not a good approach please let me know.
It has one util function for downloading and saving a card image in our local media folder. After that, a serializer Mixin returns the local URL for the requested image in every serializer where the card image is needed.

### DIFF
--- a/backend/yugioh/utils.py
+++ b/backend/yugioh/utils.py
@@ -1,0 +1,38 @@
+import os
+
+import requests
+from django.conf import settings
+
+
+def fetch_and_save_image(image_url):
+
+    external_image_url = image_url
+
+    local_image_name = image_url.split('/')[-1]
+
+    media_path = os.path.join(settings.MEDIA_URL, 'yugioh_card_images')
+
+    image_file_name = os.path.join(media_path, local_image_name)
+
+    headers = {"User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0",
+               "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
+               "Accept-Language": "en-US,en;q=0.9"
+               }
+
+    if not os.path.exists(media_path):
+        os.makedirs(media_path)
+
+    if not os.path.exists(image_file_name):
+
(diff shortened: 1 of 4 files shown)


backend/yugioh/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    if not os.path.exists(image_file_name):
+
+        image_response = requests.get(external_image_url, headers=headers, stream=True)
+
+        if image_response.status_code == 200:
+
+            with open(image_file_name, 'wb') as handler:
+                for chunk in image_response.iter_content(chunk_size=8192):
+                    handler.write(chunk)
+        else:
```
julkascript, line 34: It's better to handle else cases properly, or at least raise a warning/error (503).

backend/yugioh/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    external_image_url = image_url
+
+    local_image_name = image_url.split('/')[-1]
+
+    media_path = os.path.join(settings.MEDIA_ROOT, 'yugioh_card_images')
+
+    image_file_name = os.path.join(media_path, local_image_name)
+
+    headers = {"User-Agent": "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0",
```
julkascript, line 17: Are you sure you need the headers?

milensski: > Are you sure you need the headers? What is their purpose in this case? Also, if yes, move them inside the if (before the request).

i was not sure, it actually works without them, and reformated the code.

Thanks.

julkascript:
Fix conflicts

backend/yugioh/utils.py
milensski: > It's better to handle else cases properly, or at least raise a warning/error (503).

Now it should be ok -
I have made an exception to be handled properly, still returning the external image link while creating a log in the console for the exception.

dinocom33 (approved):

Issue #117, opened 2024-06-09
Implement Stats Tracker


julkascript:
Backend requirements:
- Track and provide in the API:
  - How many times a user has bought cards (checked out from the Checkout) - "Purchases"
  - How many sales the user has done (orders). - "Sales"
  - Sales this month - "Sales This Month"
  - Average rating from reviews - "Seller Rating"
  - How many times has the seller rejected a sale. - "Rejection Rate"
  - How many times has the seller not sent a sale. - "Miss Rate"

Frontend requirements:
- Display the correct information in the User Profile section.

![image](https://github.com/julkascript/cardflow/assets/59143945/09356c8c-eca6-406b-8c52-a29a11097f06)

milensski:
I have a question about the last point:

- > How many times has the seller not sent a sale. - "Miss Rate"

Currently, for an order status, we only save if it's rejected for whatever the reason is, both from the seller and buyer,
I can figure out a way to check how many times the seller rejected but I am not sure how to check with the current setup if he missed to sent, since there is only a status rejected.

milensski:
Can you please check my logic and advise if you agree that we can count it like that,

- How many times has the seller rejected a sale. - "Rejection Rate" -> a rejection rate will be counted if the seller after status ordered immediately rejected the order.

- How many times has the seller not sent a sale. - "Miss Rate" -> miss rate will be counted if the seller sent the order but the buyer marked it as not received.

julkascript:
- Yes
- No. There is the possibility that a seller can just literally not send a package (ghost the buyer). So if the seller does not send the package in 7 days, we should mark the order as "not sent".

milensski:
I understand, that makes sense now. Thanks! 🙌

PR #125, opened 2024-06-29
Stats tracker Backend and Frontend

milensski:
This PR resolves issue #117

For the Backend:
I integrated the stats to be returned in the request with the user profile data since this is public info and can be returned at once.
Let me know what you think of this approach I will reformat it if it needs to be separated. 

For the Fronted:
Since everything was set up on the frontend I only added the data to their fields. I am not familiar with React so please let me know if something should be done differently.


RyotaMitaraiWeb (approved):
regarding the front end code, you will want to configure a linter in your IDE so that it can warn you about stuff like missing semicolons, unnecessary spaces, and other rules we have set up, since stuff like this was committed in your code. I have fixed those myself to save us the trouble.

Outside of this, the trackers seem to work—it even calculates stuff like seller rating from orders made before the feature was implemented. The only thing I didn't test was the miss rate, since none of the orders I have on my end are marked as not sent, but it seems like the backend code handles the calculation properly

milensski:
Thanks, I will note that. 👌

dinocom33 (approved):
LGTM

julkascript:
Sorry, conflicts appeared after some merges @milensski

milensski:
> Sorry, conflicts appeared after some merges @milensski

Should be ok now.

frontend/src/services/user/types.ts
```diff
@@ -55,4 +55,12 @@ export type UserAccountLoader = {
 export type PublicUserInfo = {
   username: string;
   avatar: string;
+  stats: {
+    purchases: number;
+    sales: number;
+    sales_this_month: number;
+    seller_rating: number | string;
```
julkascript, line 62: Why can this be either a number or a string?

julkascript:
Other than the small comment, LGTM

julkascript (approved):

frontend/src/services/user/types.ts
milensski: it comes from the backend as string because i am formating it there and sending it but I kept the number if I need to change it. If this is no ok, I can change it to number and then format it in the frontend ?

Issue #128, opened 2024-06-30
Beta Note


julkascript:
Add a note in the About page, which informs the user that the app is in a beta stage.

Text: The app is still in beta, some features may not be available! If you find any bugs, don’t hesitate to report them.

[Figma design](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=2653-133&t=txQ0JRlo1Dw2NpLP-0)

Alperadoss:
Hi, can you assign me to this issue? I'd like to do it

julkascript:
@Alperadoss Hey, welcome, please feel free to pick up the ticket! Let me know if you need any assistance.

PR #134, opened 2024-07-01
beta alert  div coded

Alperadoss:
PR for issue : https://github.com/julkascript/cardflow/issues/128

I  translated to Bulgarian language with Google Translate, and am not sure if it is correct translation. Can you please check?

Screenshot 2024-07-01 at 17 41 23


frontend/src/translations/bg/about.json
```diff
@@ -5,6 +5,10 @@
   },
   "main": {
     "title": "За нас",
+    "betaWarning": "Приложението все още е в бета версия! Ако откриете грешки, не се колебайте да ги докладвате.",
```
julkascript, line 8: Is this ‘betaWarning’ translation used? If not, remove it.

frontend/src/components/about/BetaAlertDiv.tsx
```diff
@@ -0,0 +1,58 @@
+import { useTranslation } from 'react-i18next';
+import Alert from '@mui/material/Alert';
+import CloseIcon from '@mui/icons-material/Close';
+import { Button } from '@mui/material';
+
+type BetaAlertDivProps = {
+  setAlertVisible: (isVisible: boolean) => void;
+};
+
+const BetaAlertDiv: React.FC<BetaAlertDivProps> = ({ setAlertVisible }) => {
```
julkascript, line 10: I’d recommend you to name the component simply ‘BetaAlert’. To me ‘div’ seems redundant.

Also, I think this alert does not need to be in a separate component. But I also want to hear what @RyotaMitaraiWeb thinks….?

frontend/src/components/about/BetaAlertDiv.tsx
```diff
@@ -0,0 +1,58 @@
...
+        borderRadius: '8px',
+        alignItems: 'center',
+        '.MuiAlert-icon': {
+          fontSize: 30,
+        },
+      }}
+    >
+      {t('main.betaWarningFirstPart')}{' '}
+      <a
+        href="http://localhost:5173/about/contact"
```
julkascript, line 47: This is not the correct way to redirect. We don’t redirect to localhost because we need to be compatible with all environments where we deploy. Please take a look at other files how we do it.

julkascript:
Thanks a ton for your contribution! I wrote some comments you can address. @RyotaMitaraiWeb may also have some remarks. 

Overall you’re in the right direction!

RyotaMitaraiWeb:
- the translation seems ok at first glance, but I recommend looking into the [Trans component](https://react.i18next.com/latest/trans-component) to interpolate the hyperlink, rather than breaking it in three different translations ([see an example of that in the repository](https://github.com/julkascript/cardflow/blob/33ef4511baf1fe001dc3807f6bcd128e8fc3b3a1/frontend/src/components/auth/AuthForm.tsx#L211))
- use React Router's ``Link`` component instead of a normal ``a`` tag.
- prefer using TailwindCSS class names instead of ``sx`` when possible. For example, you can use ``w-5/6`` instead of a manual ``width`` (and do this for all other ``sx`` stylings where possible). Use ``sx`` only when Tailwind class names don't work. Refer to the [TailwindCSS documentation](https://tailwindcss.com/docs/installation) if you aren't familiar with the available class names (use the menu on the left for navigation)
- @julkascript has to weigh in about this, but I feel like being able to hide the alert seems pointless, since navigating back to the page will restore it anyways. Furthermore, I don't think users will be spending enough time on the page to the point that having a persistent alert is going to cause them much harm.

frontend/src/components/about/BetaAlertDiv.tsx
RyotaMitaraiWeb: feels like an overkill to me, personally, it's not really abstracting or encapsulating anything that the component already doesn't.

frontend/src/translations/bg/about.json
Alperadoss: i refactored with Trans as RyotaMitaraiWeb advised and removed unused translations.

frontend/src/components/about/BetaAlertDiv.tsx
Alperadoss: I moved it back into main component About.tsx.

frontend/src/components/about/BetaAlertDiv.tsx
Alperadoss: I'm embarrassed about local link, sorry for it. I changed it to href="/about/contact" as you use in repo that way.

Alperadoss:
Thanks for directing @julkascript  and @RyotaMitaraiWeb .
I got one unresolved issue. When i apply tailwindCSS in Alert it doesn't apply in anyway. I tried to wrap with a div and apply tailwind to that but it didn't work too. I also had a look at other places in repo where "className=" is used but didn't see sth to solve this issue. Can you please help with this one?

Dear @julkascript if you also advise as @RyotaMitaraiWeb  I can remove useState and remove hiding alert div.

julkascript:
> @julkascript has to weigh in about this, but I feel like being able to hide the alert seems pointless, since navigating back to the page will restore it anyways.

I agree. Handling state for the Alert is unnecessary. You can remove the state and let the Alert be there all the time.

@Alperadoss can you please give a specific code example of what you tried with tailwind, so we can help?

Alperadoss:
I removed useState. Tailwind CSS code example is below. 

`      <Alert
        id="closable-alert"
        severity="info"
        className="text-black text-lg w-5/6 mx-auto my-4 bg-[#E5F6FD] flex font-light border rounded-lg items-center"
      >
        <Trans i18nKey="betaWarning">
          {t('main.betaWarning')}
          <Link href="/about/contact" />
        </Trans>
      </Alert>`

RyotaMitaraiWeb:
Managed to pull the changes locally

What seems to trouble @Alperadoss based on what I see on my end is the fact that the font class names do not apply:

alert

If we care enough about matching the font exactly as on the wireframe (I personally think MUI's default is fine, but I will let others judge), you need to wrap up the ``Trans`` component in another tag (e.g. a ``<p>``) and apply the classes there

julkascript:
Please stick to the default font (MUI’s default). The wireframe is just an example, sorry if it brings confusion.

RyotaMitaraiWeb:
In that case, only remove the classes that visibly don't do anything (as to not litter the component with useless class names) and I will approve

Alperadoss:
I removed the classes which doesn't make change regarding @julkascript 's message. 
Screenshot 2024-07-03 at 18 00 33

RyotaMitaraiWeb (approved):
looks good to me

frontend/src/pages/about/About.tsx
```diff
@@ -43,10 +43,21 @@ function About(): JSX.Element {
       logoColor: theme.palette.error.light,
     },
   ];
+
   return (
     <section className="bg-[#F5F5F5] min-h-[100vh] pb-4">
       <CardflowTabs />
-      <PageHeader heading={t('main.title')} />
+      <PageHeader heading={t('main.title')} />{' '}
```
julkascript, line 50: Remove unnecessary `{' '}`

julkascript:
This change was planned for v1.4. However, since it's done, it looks good and it's a small change - I'm happy to actually include it in v1.3, once my latest comment gets addressed.

PR #135, opened 2024-07-02
Dynamic currency

dinocom33:
This PR adds dynamic currencies of the user's choice. All prices will be displayed in the currency selected by the user (if authenticated). If the user is not authenticated, all prices will be displayed in BGN. In the database, the prices will be recorded in BGN regardless of the currency chosen by the user.

I'm not sure I covered all the edge cases, so if you find a problem please let me know.


julkascript:
Looks good. Please add endpoint, which allows the user to update their currency preference. This endpoint will be accessed from the frontend. Make sure users can only update their own preferences.

dinocom33:
The currency preference can be updated trough /api/accounts/user/{id} endpoint with PUT or PATCH request and these operations are allowed for current logged user only. Do we need а separate endpoint for changing the currency?

julkascript:
Yes, my bad, totally missed it. All good!

julkascript (approved):

PR #136, opened 2024-07-04
Not received

dinocom33:
This PR adds the Not Received feature. The buyer can mark the order in this way only after 10 days have passed from the date of placing the order. All order items from this order are restored in this operation.


julkascript (approved):

backend/order/serializers.py
```diff
@@ -72,6 +74,22 @@ def update(self, instance, validated_data):
                     order_listing.is_sold = False
                     order_listing.save()
 
+            if validated_data['status'] == 'not_received':
+                # Check if 10 days have passed since the order was placed
+                if instance.status_history.exists():
+                    order_placed_date = instance.status_history.last().timestamp.date()
+                    print(order_placed_date)
```
julkascript, line 81: Leftover print

PR #137, opened 2024-07-05
Introduce dynamic currencies to the front end

### LINKED ISSUE #113: Dynamic currency
General requirements:
- There should be 2 available currencies:
  - BGN
  - EUR
- The default currency should be BGN.

Frontend requirements:
- The currency can be changed from Account Settings.
- There is a new section for the default currency setting.
- At all places in the app where prices are displayed, display the currency after the price. E.g.,     12.23 лв., 55.12 EUR
- The dollar sign ($) is removed for now. In the future if we add dollars as currency, we can bring them back.

Backend requirements:
- Handle the user's choice in the backend for persistency.
- Price is stored as a value, which is linked to the default currency of the app. So if the user has selected BGN, and chooses 20 BGN, the stored value will be 20. If the user has selected 20 EUR, the stored value will be 39.2. 
- The API is responsible for returning the correct value, according to the user selected currency preference. 

Exchange rates:
1 EUR == 1.96 BGN
1 BGN == 0.51 EUR

[Design in Figma](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=624-28&t=iV1mkpkzoXyvVsjC-0)

RyotaMitaraiWeb:
Closes #113 

This PR allows the user to change their currency on the front end and also displays prices in their preferred currencies. There are several things to note:

- the frontend does not do any conversions whatsoever, every price displayed is taken directly from the backend.
- the above results in a few bugs that will need to be addressed (planning on opening new issues for them):
  - the ``total_price`` property of the shopping cart always returns the price as BGN
  - when the user enters a price in EUR when creating or editing a listing, it is possible for the price to be slightly lowered for EUR users due to the fact that all prices are internally converted to BGN (for example, 60 EUR becomes 59.98 in the eyes of the user)
- in the profile settings page, each currency option also includes the full name of the currency for better clarity.
- when the currency is updated, the user state is updated, although the access token containing the old currency remains. I am taking advantage of the fact that our app generates a new access token every time the app loads, preventing this from becoming an issue. I mention this in case we decide to alter that behavior.
- in the new / edit listing pages, the preferred currency is placed at the end of the price field so that the user knows what currency they are submitting their listing in
- for the English side of the platform, BGN prices are still displayed as "лв.", I am not expecting this to cause issues.


julkascript:
Neat! 🔥 

Could not test this because I'm away from my personal PC for a few days, but I reviewed the PR. Implementation-wise looks good. I approve. 

About the bugs - please open tickets when you have capacity and we'll be able to prioritize.

julkascript (approved):

PR #142, opened 2024-07-07
Make conversions precise, fix shopping cart currency bug

### LINKED ISSUE #139: Make creating and editing listings with euro as preferred currency consistent
**Issue**
Because of the fact that all prices are internally converted to BGN, submitting prices in euro results in them being different from what the user submitted. This obviously isn't particularly desirable, so it should be fixed.

**How to reproduce**
1. Set ``EUR`` as your preferred currency (this can be done in the profile settings page)
2. Go to the new listing page
3. Select a card, quantity, and condition and input 60 EUR as price
4. Create the card
5. Notice that the displayed price is *not* 60 EUR (it should be around 59.98 EUR, I think)

(I also managed to reproduce this via the Swagger client)

**Potential solutions**
(there could be a better way to implement this, this is merely a suggestion)
1. create a new column for the listings called ``submitted_currency`` or something similar, which is set to the submitter's preferred currency at the time of creating or editing the listing
2. If we want the backend to be doing the conversions, the backend checks the request's preferred currency and compares it to ``submitted_currency``. If those match, return the stored price directly, if not, convert.
3. If we want the frontend to be doing the conversions, the backend includes the ``submitted_currency`` in the response and lets the frontend handle the conversion in a similar algorithm as the above

(this has the added benefit of allowing us to automatically migrate all BGN prices to EUR in the event Bulgaria enters the euro area).

### LINKED ISSUE #140: Shopping cart's ``total_price`` returns the value only in BGN
**Issue**
Regardless of your preferred currency, the ``total_price`` property of the shopping cart items is returned in BGN for the ``/api/cart/items/`` endpoint (when performing GET requests).

**How to reproduce**
1. Set ``EUR`` as your preferred currency (this can be done in the profile settings page)
2. Add an item to the shopping cart, keeping its BGN price in mind
3. Send a request to get all your shopping cart items and note that ``total_price`` for each item is displayed in BGN
(all of this can also be reproduced on the Swagger client)

julkascript:
Closes #140 
Closes #139

@RyotaMitaraiWeb I decided that it'd be easier to just make the conversions more precise, instead of reworking the way we save the prices. etc. We may definitely consider this in the future, but for now, I just want to get the product out for public beta testing with fewer bugs.

### DIFF
--- a/backend/cart/models.py
+++ b/backend/cart/models.py
@@ -27,6 +27,8 @@ class ShoppingCartItem(models.Model):
 
     @property
     def total_price(self) -> float:
+        if self.cart.user.is_authenticated and self.cart.user.currency_preference == 'EUR':
+            return self.quantity * round(self.listing.price * 0.511292, 2)
         return self.quantity * self.listing.price
 
     def __str__(self):
--- a/backend/listing/serializers.py
+++ b/backend/listing/serializers.py
@@ -23,7 +23,7 @@ class ListingSerializer(serializers.ModelSerializer):
         if request:
             user = request.user
             if user.is_authenticated and user.currency_preference == 'EUR':
-                representation['price'] = round(instance.price * 0.51, 2)
+                representation['price'] = round(instance.price * 0.511292, 2)
         return representation
 
     def create(self, validated_data):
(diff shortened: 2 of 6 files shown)


RyotaMitaraiWeb (approved):
looks good to me

(values used to test: 60 EUR, 30.63 EUR, 4.99 EUR, 8 BGN)

PR #144, opened 2024-07-08
Replace new listing search with an autocomplete

### LINKED ISSUE #141: Creating a listing does not allow you to see all sets
If a card has many set appearances (CardInSet instances), the user cannot see all of them when creating a listing.

As it can be seen on the picture, there are 47 more sets/rarities to choose from. However, the user can only choose from the first few. If they decide to click on the "Show more" - it will take them to a dead end - the explore page.

image

Therefore, we should replace the search with an autocomplete field that can easily solve this problem ([MUI Autocomplete](https://mui.com/material-ui/react-autocomplete/)).

The autocomplete should contain the same data as the search did.

RyotaMitaraiWeb:
Closes #141 

The new listing page now has an autocomplete. Once the user types in something, a debounced request (debounced by a half a second) will make a search request and populate the menu options below it. Clicking any of those will redirect you to the create page for the respective card

Notes:
- although the options are rendered as hyperlinks, there is still a function that triggers when an option is selected that performs a programmatic navigation. The reason for that is to support keyboard navigation as well (as you can cycle through the options with arrow keys and the Home and End keys, as well as select an option with Enter), in which case hyperlinks aren't enough. This also has the benefit of things like URL preview (at the bottom of most browsers) and ability to open the page in a new tab.
- Because the search endpoint is paginated, the list will never display more than ten items. This is something that should probably be fixed before we release a stable version.
- Cards with particularly long names will require horizontal scrolling. The only (reasonably achievable) way to fix this is to make the search field wider.

### DIFF
--- a/frontend/src/components/sellListing/NewSellListingBody.tsx
+++ b/frontend/src/components/sellListing/NewSellListingBody.tsx
@@ -4,7 +4,7 @@ import { YugiohCardInSet } from '../../services/yugioh/types';
 import { useDebounce } from '../../util/useDebounce';
 import { yugiohService } from '../../services/yugioh/yugiohService';
 import { useToast } from '../../util/useToast';
-import { useNavigate } from 'react-router-dom';
+import { Link, useNavigate } from 'react-router-dom';
 import CameraAltIcon from '@mui/icons-material/CameraAlt';
 
 type SearchResults = {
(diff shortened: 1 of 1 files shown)


julkascript:
Nice, looks good overall! 

Two things, though:
1. I get a "key" error in the console when react tries to render the items. Maybe we need to take a look.
2. (If not too big of a hustle) Please put the magnifying glass icon in the beginning

RyotaMitaraiWeb:
I couldn't replicate the issue with keys, I would need more details to reproduce it

julkascript:
I recorded a gif for you:

err

Also, here is the error log:
```
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, tabIndex: ..., role: ..., id: ..., onMouseMove: ..., onClick: ..., onTouchStart: ..., data-option-index: ..., aria-disabled: ..., aria-selected: ..., className: ..., children: ...};
  <ForwardRef(MenuItem5) {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {tabIndex: ..., role: ..., id: ..., onMouseMove: ..., onClick: ..., onTouchStart: ..., data-option-index: ..., aria-disabled: ..., aria-selected: ..., className: ..., children: ...};
  <ForwardRef(MenuItem5) key={someKey} {...props} />
    at Autocomplete2 (
```

RyotaMitaraiWeb:
So I didn't manage to reproduce the error in the console, despite also trying it on Brave myself. However, I think I figured out what's going on.

From [what I am reading](https://github.com/mui/material-ui/issues/39833), the ``props`` value passed to the render prop actually has a property known as ``key`` (which I subsequently confirmed to be true). When I was spreading the props directly onto the MenuItem component, it passed its own ``key`` to it and that presumably caused React to complain. I have committed a change to remove the ``key`` from the props (if this works, this might be the first time I've ever seen ``delete`` in a real-world scenario 👀), but I need a confirmation that it solves the problem because, again, I never managed to reproduce the error in my own browsers (despite doing the same things as in the GIF).

julkascript:
Yep, correct, well done! LGTM!

PR #150, opened 2024-07-14
Adjust order modal sequence and split off some of  the layout

### LINKED ISSUE #132: Swap Order Modal information sequence
To make the UI smoother, swap the table and summary places.

Also, align the "|" separators of the Summary and History sections to be on the same line.

image

RyotaMitaraiWeb:
Closes #132 

This PR does three things:
- changes the order of the layout
- makes the separators on the desktop layout correctly aligned
- separates some of the elements in their own components. This was done to make it easier for me to finish this.

I have kept the previous UI mostly intact (e.g. the feedback section is still centered in the middle of its own part), though there may be some minor changes here and there

### DIFF
--- a/frontend/src/components/orders/ordersModal/OrdersModal.tsx
+++ b/frontend/src/components/orders/ordersModal/OrdersModal.tsx
@@ -259,7 +259,7 @@ function OrdersModal(props: OrdersModalProps): JSX.Element {
           </section>
           <Divider className="hidden lg:block place-self-center" orientation="vertical" flexItem />
           <Divider className="block lg:hidden" flexItem />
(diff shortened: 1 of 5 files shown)


julkascript:
Great job, as always! Sorry to change a design/requirement last minute, but it's something I also missed. Can we also align the Feedback to be at the same vertical level as the Shipping Details? It just looks a bit funky right now.

image

RyotaMitaraiWeb:
No biggie, this is done

(fun fact: this is how it actually looked like when I switched the layout to a grid, I had to explicitly center it)

julkascript (approved):

julkascript:
Yep, exactly, gj!

Issue #153, opened 2024-07-21
Implement "I Would Trade This" checkbox


julkascript:
Implement the "I would consider trade offers for this listing" checkbox at listing edit and creation time. The value of this checkbox will be used in the Listings screen to show the trading intent of the user. Basically, if the user marked that they would consider trade offers, this would give a better indication to other users if trade offers for this listings will be seriously considered or not.

Note: This value does not block users from requesting trade offers with that listing, it just shows that the user does not want to trade it.

Backend requirements:
- This field can be named `would_trade` on the backend (or however you prefer), as long as it's a bool.
- This field should be returned with listings GET requests

[Figma design](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=212-209&t=7g13gQ10hJyn6Oem-0)

PR #158, opened 2024-07-22
Implement "I Would Trade This" checkbox

dinocom33:
This PR corresponds to the backend part of issue #153. It adds a new field named ```is_tradable``` (i chose this name to be consistent with the other ones like ```is_sold``` and ```is_listed```.
This field is a bool and it returns with listings GET requests.
It has also been added to search filters as well as in the admin page.


julkascript:
Looks good, however let's rename `is_tradeable` to `is_considering_trading` because `is_tradeable` leads you to believe if the value is false, the card should not be tradeable. which is misleading.

dinocom33:
The field's name was changed to ```is_trade_considered```

julkascript (approved):

PR #161, opened 2024-07-25
Implement trade checkbox on the frontend

### LINKED ISSUE #153: Implement "I Would Trade This" checkbox

RyotaMitaraiWeb:
Closes #153 

This PR adds the checkbox from the Figma wireframes in a functional condition. On viewports < 1024px, the checkbox is below the three fields due to there being no room to put it horizontally

### DIFF
--- a/frontend/src/components/sellListing/ListingForm.tsx
+++ b/frontend/src/components/sellListing/ListingForm.tsx
@@ -87,6 +87,7 @@ function ListingForm(props: ListingFormProps) {
           rarity_code: '',
         },
       },
+      is_trade_considered: false,
     },
   );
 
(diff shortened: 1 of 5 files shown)


julkascript:
🔥

julkascript (approved):

Issue #116, opened 2024-06-09
(Draft) Trade


julkascript:
General requirements:
- A user (Requester) can request trade to another user (Responder)
- The Requester can select what they require from a predefined list, which is generated from the Responder's listings. In other words, Requesters can only request what Responders have listed in their "inventory".
- The Responder can Accept, Decline, or Negotiate
  - The Negotiation is like a "counter-offer". The Responder and the requester switch places. This can be done multiple times.
-  Users can also additional options, which for now can only be "cash". For example, a user can trade "card X" for "card Y" and 20 BGN.

RyotaMitaraiWeb:
I am seeking a clarification on some behavior, because the way it's implemented on the backend and the way it's portrayed in the wireframes is confusing me:

- reading the requirements and the wireframes (in particular the chat ones), I assume that one should be able to change the other user's listings. In addition, when initiating a trade offer, you would obviously need to do that as well so that the other user knows what you want.
- However, on the backend, sending a ``PATCH`` request to the trade endpoint gives me a 400 error when I try to modify the other's listings, stating that I am not allowed to change the other user's listings
- On top of this, the update *actually* succeeds despite the 400 error (this one feels like a bug)

Am I understanding / doing something wrong? If so, what would be the correct behavior / course of action in that case?

julkascript:
> I am seeking a clarification on some behavior, because the way it's implemented on the backend and the way it's portrayed in the wireframes is confusing me:
> 
> * reading the requirements and the wireframes (in particular the chat ones), I assume that one should be able to change the other user's listings. In addition, when initiating a trade offer, you would obviously need to do that as well so that the other user knows what you want.
> * However, on the backend, sending a `PATCH` request to the trade endpoint gives me a 400 error when I try to modify the other's listings, stating that I am not allowed to change the other user's listings
> * On top of this, the update _actually_ succeeds despite the 400 error (this one feels like a bug)
> 
> Am I understanding / doing something wrong? If so, what would be the correct behavior / course of action in that case?

Yes, you are correct. A PATCH endpoint should allow changes of the listings on both sides. However, the listings should still be owned by the other user, you should not be able to select listings of other users. Maybe it’s a bug. @dinocom33 can you double-check?

dinocom33:
If you can fix it, do it or raise an issue. Which non registered actions do you mean?

RyotaMitaraiWeb:
> If you can fix it, do it or raise an issue. Which non registered actions do you mean?

[...]

PR #164, opened 2024-08-06
Trade

dinocom33:
This PR represents Trade functionality(issue #116 ).
When a user initiates a new trade, he becomes the ```initiator``` and his ```initiator_decision``` is automatically filled in as accepted and the status of the trade becomes ```Negotiate```.
If one of the users rejects the trade, it automatically becomes ```Rejected```, the trade is closed and cannot be changed anymore.
If they both accept the trade, it's becomes ```Accepted```.
Users can only trade their listings which are marked as ```is_listed```.


julkascript:
I tried Trade create request with body:
```json
{
    "initiator_listing": [
        1
    ],
    "recipient_listing": [
        2
    ],
    "initiator_cash": 5,
    "recipient_cash": 0,
    "initiator_decision": "pending",
    "recipient_decision": "pending"
}
```

but I get error: `"You can not trade with recipient's listing ids: [2]."`. Id 2 is a valid target, because that is another user. I am currently logged in as user 1, and I am using its token. I get the same error when I try with other ids.

julkascript:
Also, I think it will be easier if we enforce the `recipient id` (currently we obtain it like `recipient = recipient_listing[0].user`, which is easy to get around).  Then check if all listings you are trying to trade are really of this user and this user only.

julkascript:
Otherwise looks good!

dinocom33:
Fixed. Now the POST request looks like bellow:
```
{
    "recipient": 4,
    "initiator_listing": [
        3, 4
    ],
    "recipient_listing": [
        14
    ],
    "initiator_cash": 5,
    "recipient_cash": 0,
    "initiator_decision": "pending",
    "recipient_decision": "pending"
}
```

julkascript (approved):

Issue #157, opened 2024-07-21
Implement Trade Chat


julkascript:
The Trade Chat is a chat between the two traders. It should help the users arrange the trade details (shipment, delivery, etc.).

- From a technical POV, each message will have a sender. The sender can be any of the two users, or the "System" because there are sometimes system messages.
- The "Click to expand offer..." button opens the current offer modal.
- The chat can continue even after a deal is reached.
- A new chat is started per trade session.

Note: To simplify, the chat does not need to be responsive. Users can refresh to obtain the latest messages.

[Figma design](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=2779-179&t=7g13gQ10hJyn6Oem-0)

julkascript:
Some more pointers for this ticket. 

The system will have to send a chat message when:
- a deal is reached (both sides accept)
- on rejection (deal is rejected)
- one of the sides negotiates
- on initial offer

PR #168, opened 2024-08-21
Implement trade modal on the frontend + a few changes to the backend

RyotaMitaraiWeb:
Related to #116 

**Note 1:**  the current implementation has a few bugs / unwanted side effects, but the backend will have to be tweaked slightly to fix them. Read below for more info.
**Note 2:** any UI below the modal will be implemented with #157 in the future

This PR implements the following:
- the trade modal with a few caveats
- a few changes to the backend to better suit this PR's needs
- a change to the search users page to make the trade button functional (the button initiates an empty offer and it is up to the sides to be more specific as to what they want)

**Modal**
- The modal allows the user to select listings and cash options on both sides, as well as accept, reject, and request offers.
- The modal, for the purposes of this PR, can currently be opened only by navigating to the trade offer's page ``/trade/{id}``

**Modal (listing search)**
- the listings search is similar to the autocomplete search on the New Listing page
- in addition to only searching visible listings, the results also won't display any listings that are already in the offer (this is a client-side thing only, the duplicate listing is still fetched)
- the search is disabled when the trade is no longer active
- Note that the search is paginated, so only 10 items at best will be displayed. In addition, because of the second bullet, if there are duplicates with the current offer, the actual maximum amount will be less than 10.

**Modal (cash field)**
- the cash option works as you might expect it to, the only notable thing to note is that changes to the field are debounced by a half a second so that the frequent updates to the page don't annoy the user
- Like the listing search, it is disabled when the trade offer has concluded

**Modal (offer on the right)**
- for each selected listing, an image, alongside a remove button and chips with small info, is displayed on the right
- the remove button removes the listing from the offer and is visible only when the trade is active.
- if a cash option is specified, there will also be a box that displays the cash option with the currently logged in user's preferred currency
- - **Note:** the frontend does not do any conversion of the currencies, so this will lead to inconsistencies, might be worth dealing with this on the backend
- If no cash option is specified, then the section will display an empty box instead
- In both cases, the box is always at the end of the list, regardless of whether it was added before or after a specific listing.
- Note that changes do not get saved automatically, the user has to explicitly request them in order to save them.

**Modal (buttons)**
- The close button is always visible and closes the modal (without saving any changes)
- The request button is visible only when the trade offer has been changed (regardless of which side the change has been made to). For the purposes of this PR, an offer is considered different if any of the cash options are different from the original or if a listing was added or removed from any of the sides. Clicking the request button invokes a "negotiation" on the backend
- The accept and decline buttons are visible only when there are no changes to the offer. Clicking them invoke the respective events on the backend
- All buttons (minus the close one) are hidden if the trade is concluded and disabled if any of the sides is "empty" (aka the side has no cash option configured *and* no listings offered / requested)

**Backend changes**
Feel free to let me know if any of the changes are problematic
- I added a new filter to search listings by their card names, as I needed this for the listing search feature
- I added a new user serializer that includes the avatar (which I needed for the section on the left). I opted for this to prevent potential side effects in other parts of the app.
- this one may be me having no idea what I am doing, but there were a bunch of actions (negotiate, accept, reject, and counter_offer) that I could not make requests to and so registered them explicitly so I can work with them, let me know if I wasn't supposed to do this
- removed validations that prevented an initiator from sending ``recipient_listing`` and vice-versa. Based on [this comment](https://github.com/julkascript/cardflow/issues/116#issuecomment-2297273568), both sides should be able to change each other's listings.
- the custom validator for ``initiator_listing`` had a bug; because it used the logged in user's ID, it made it impossible for the recipient to request listings from them. To fix this, I included the initiator ID in the serializer (alongside the recipient ID) and use that instead to check for proper ownership of the listings. As far as I have tested, users are still unable to make changes to trades that aren't theirs, but let me know if I introduced a regression like this.

**Known issues / things that I will want to implement at some point / potential issues**
- currently, the buttons do not take into account the users' trade decisions. This leads to a few scenarios that could be seriously abused. For example, if one user has accepted the offer, but the other changes it and then accepts it, then the trade is concluded. The reason this has not been addressed for now is because I will need the backend to perform the needed changes to the trade decisions (since an initiator cannot change the recipient's decision and vice-versa, for good reasons; we don't want users to abuse this); for example, if the recipient changes the offer, the recipient decision should become "accepted" and the initiator decision should become "pending"
- I am not quite sure when both ``negotiate`` and ``counter_offer`` are supposed to be used in the trade flow, so I had to do some guesswork here. Currently, counter offers aren't used at all on the frontend. If I need to change this, please let me know

Overall, there might be some issues with the implementation because I myself got confused at times as to how a feature is supposed to work, so please let me know what changes will be needed

(Side note: introducing backend changes in such amounts in a branch labeled as "frontend" probably isn't the most appropriate, but I was too deep into this by the time I realized it D:)


julkascript:
Sheesh, big PR. Great job overall, I can't possibly review everything but it looks okay, I also did some testing. I would like @dinocom33 to have a say on the backend part.

I accept the known issues and potential issues. We can fix them on the next iterations (backend as well). For a first iteration, this is a solid start! 🔥

julkascript (approved):

dinocom33 (approved):
As for the backend part, the changes look ok to me. I tested the operations related to the trade (not sure if I tested everything) and they work as expected. We will certainly make changes to the backend in the future, but for this PR I think they are enough. Regarding the frontend, I think there should be something like show all trades and show all listings, so that the user can choose by browsing, because right now to buy or trade a card, I need to know the exact username and card name.

PR #169, opened 2024-08-25
Implement "My Trades" on the frontend

### LINKED ISSUE #156: Implement My Trades
Implement an overview with all trades. 

- Trades should have a unique `TR-<number>` id.
- Trades should have a current status:
  - Awaiting their response
  - Awaiting your response
  - Deal reached (Completed)
  - Rejected
- Clicking on the trade takes the user to the trade page.

The "New Trade" button should take the user to the [User Search](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=2824-100&t=7g13gQ10hJyn6Oem-0).

[Figma design](https://www.figma.com/design/GpsRZnN69f29pOCkiciWD4/Cardflow?node-id=2777-42&t=7g13gQ10hJyn6Oem-0)

RyotaMitaraiWeb:
Closes #156 

This PR does a few things, outside of implementing the page:
- moves some of the logic in the ``/sell`` page into reusable components and hooks so that they can be reused here and other places if we want similar UI for future tasks.
- moves the ``CardflowTabs`` component in its own standalone component (as well as making the trade button functional).

**New hook and component**
A new hook, ``useSelect``, has been introduced to abstract away the logic related to the checkboxes in the first cells of the table rows.

A new component, ``ActionsMenu``, has been introduced, which is basically the menu with buttons found at the bottom of the page. Through a configuration array, you can generate it without having to write markup at all, which should make it easier to reuse if we need another page like this in the future

**My Trades page**
The page works similarly to the ``/sell`` page; once the user navigates in, it retrieves a paginated list of trades and renders them.

An important distinction, though, is that the checkbox on the left of each trade is disabled if the trade has ended. In addition, when pressing an "All" action, the client won't bother sending requests for those trades (since it'd be pointless). Other parts of the UI, likewise, take into account the disabled status (e.g. when counting the amount of selected trades)

**Note:** it seems like the ordering on the backend is inconsistent; when I update a trade's status, the trades aren't in the same order after I refetch them. Not gamebreaking, but probably worth adding to the backlog at least


julkascript:
LGTM!

About the inconsistent ordering - I'll need to think about that because it also looks like a feature :D because it would always put the item that you updated first.

julkascript (approved):

PR #174, opened 2024-11-06
Implement (partially) trade chat on the frontend

RyotaMitaraiWeb:
Addresses most of #157 

This PR adds the trade chat to the trade page, alongside the side menu that shows the current offer. There are two caveats, the latter of which is probably major:

- the middle button in the side menu merely opens the modal, since I wasn't quite sure what it's supposed to be
- currently, the chat only sends user-generated messages. I haven't had an opportunity to test it, but I suspect the expectation is that the client should generate system messages. If that's the case, it will probably be better to generate them on the server, rather than on the client, as this can be abused maliciously (e.g. I could send a phishing / spam link and use the system message to misleadingly prompt the user into clicking my link). If for the alpha version it's enough to generate them on the client, then I can look into implementing it that way, but for now I am leaving this as a note.

the trade chat will refresh every time the user sends a new message. It's worth noting that there is NO polling as of this moment.

(P.S.: this has been done for a few days, but lack of time + wanting to make sure I've covered as much as possible for the issue caused some delays on my end)


julkascript:
Thank you for the work, great as always! To address your comments:
> the middle button in the side menu merely opens the modal, since I wasn't quite sure what it's supposed to be
- That's totally okay, and aligns with the idea.

> currently, the chat only sends user-generated messages. I haven't had an opportunity to test it, but...
- That's okay for now. Since the idea is to get alpha testing, we are not interested in these use cases (for now). So it passes the check.

There are some bugs I found, but we won't be fixing them in this PR. I will approve for now, and we will tackle them in separate PRs.

julkascript (approved):

PR #179, opened 2024-12-12
Trade Statuses are not accurate

dinocom33:
This PR should fix trade statuses. The changes are as follows:

New endpoint - ```/api/trade/{id}/negotiate/``` which handles counteroffers. When some of the participants offers money or another listing, his status changes to ```accept``` and the status of the other one changes to ```pending```. Trade status stays ```negotiate```. 
When the counteroffer includes another listing, the user ID must also be provided in the request body, which is not necessary when offering cash.

Endpoint - ```/api/trade/{id}/accept/``` which handles manual accept. When the status of  both participants is ```accept```, trade status automatically becomes ```accepted```

Endpoint - ```/api/trade/{id}/reject/``` which handles manual reject. When one of the participant rejected the trade, trade status automatically becomes ```rejected```.
I think that if one of the two participants rejects the offer, there is no point in waiting for the other's decision.


backend/trade/serializers.py~
julkascript, line 1: Is this file intentional or a copy by mistake?

backend/trade/views.py
```diff
@@ -21,39 +21,64 @@ class TradeListingViewSet(viewsets.ModelViewSet):
...
 
-        serializer = self.get_serializer(trade, data=request.data, partial=True)
+        # Check if user modifies cash or listings
+        data = request.data
+        changed = False
+
+        if 'initiator_listing' in data or 'recipient_listing' in data or 'initiator_cash' in data or 'recipient_cash' in data:
+            changed = True
+
+        if changed:
```
julkascript, line 55: Small one, but I would delete lines 50-54 and just put line 52 for this check

julkascript:
I would say it makes more sense for these endpoints to be PATCH, because we only update a few fields, and not the whole object. Even POST works, I guess. But PUT can be a bit confusing.

backend/trade/serializers.py~
dinocom33: It was copied by mistake. I don't even have this file locally.

backend/trade/views.py
dinocom33: Done

dinocom33:
> I would say it makes more sense for these endpoints to be PATCH, because we only update a few fields, and not the whole object. Even POST works, I guess. But PUT can be a bit confusing.

Done

julkascript (approved):

